PR #91, opened 2017-06-01
Fix #40: Implement Share functionality using NSSharingService

### LINKED ISSUE #40: Make Share toolbar button work

bjtitus:
Simple implementation for sharing using NSSharingServicePicker. I just included the current link and no other text.

### DIFF
--- a/Evergreen/MainWindow/MainWindowController.swift
+++ b/Evergreen/MainWindow/MainWindowController.swift
@@ -69,6 +69,14 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 		if item.action == #selector(openArticleInBrowser(_:)) {
 			return currentLink != nil
 		}
+        
+        if item.action == #selector(showShareWindow(_:)) {
+            if let link = currentLink {
+                return URL(string: link) != nil
+            } else {
+                return false
+            }
+        }
 		
 		if item.action == #selector(nextUnread(_:)) {
 			return canGoToNextUnread()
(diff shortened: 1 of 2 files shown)


Evergreen/MainWindow/MainWindowController.swift
```diff
@@ -69,6 +69,14 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 		if item.action == #selector(openArticleInBrowser(_:)) {
 			return currentLink != nil
 		}
+        
+        if item.action == #selector(showShareWindow(_:)) {
+            if let link = currentLink {
+                return URL(string: currentLink) != nil
+            } else {
```
joeblau: Right here, you're safely unwrapping the `currentLInk` optional, but then you're never using the safely unwrapped optional.

Also, this check would be a lot simpler written as this.
1. Early termination
2. Safely use unwrapped optional

```swift
    guard let link = currentLink else {
        return false
    }
    return URL(string: link) != nil
```

Evergreen/MainWindow/MainWindowController.swift
```diff
@@ -89,6 +97,13 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 			openInBrowser(link)
 		}		
 	}
+    
+    @IBAction func showShareWindow(_ sender: AnyObject) {
+        if let link = currentLink, let url = URL(string: link) {
+            let picker = NSSharingServicePicker(items: [url])
+            picker.show(relativeTo: sender.bounds, of: sender as! NSView, preferredEdge: NSRectEdge.minY)
+        }
```
joeblau: `sender` should shouldn't be force unwrapped.  All of the parameters to the picker can be safely unwrapped before being used.

```swift
if let link = currentLink,
    let url = URL(string: link)
    let sender = sender as? NSView {
        
    let picker = NSSharingServicePicker(items: [url])
    picker.show(relativeTo: sender.bounds, of: sender , preferredEdge: NSRectEdge.minY)
}
```

bjtitus: You're absolutely right, I totally missed this somehow, thanks for taking a look! Cleaned this up and switched to using a `guard` as well.

Evergreen/MainWindow/MainWindowController.swift
bjtitus: Not sure how this `currentLink` use snuck in. My local copy has `link` being used.

I'm kind of torn on using `guard` when the statement in the scope is only a single line. I ended up changing to `guard` here and also changing to use `switch` for these cases, which I think better separates each selector path.

PR #296, opened 2018-01-07
Add very minimal support for a scripting dictionary: the getURL event

olofhellman:
Also, add an XCTestCase that can run and verify results of AppleScripts that target Evergreen.


olofhellman:
Not sure if I put the AppleEventUtils.swift and NSAppleEventDescriptor+UserRecordFields.swift files in the right place in the file hierarchy.  They currently get used in the Evergreen Tests target, but might eventually get used in the Evergreen app target as well.  Brent probably has opinions about where they belong.

Issue #411, opened 2018-08-28
buildEvergreen script — get version # from info.plist


brentsimmons:
Right now the version is specified via command line. But it should just get it from Info.plist.

Use plutil? Do I remember that name correctly? Something like that.

And then we can just call buildEvergreen and it will do the right thing.

PR #418, opened 2018-08-31
Revise the build script to determine the application version from the…

danielpunkass:
… built product, and some other niceness...


danielpunkass:
This fixes issue #411

Issue #425, opened 2018-09-02
Remove libSwift stuff from Safari Extension if possible


brentsimmons:
Having those libraries means they’re including twice in the download. If it’s possible to not have them there, then cool.

If it’s not possible, then we should file a Radar.

PR #426, opened 2018-09-02
Stop embedding Swift libraries in the Safari App Extension

danielpunkass:
Rely upon the bundled Swift libraries in the host app instead.


danielpunkass:
Fixes #425.

PR #429, opened 2018-09-04
Use query selector to better detect feed links

da2x:
Avoid downgrading HTTPS to HTTP. Introduces security and privacy risks; plus reduces performance with unnecessary redirects.


da2x:
It's not substring matching. It's matching the string "alternate" in a white space separated list. 

Here are a few examples of what the new selectors matches:

    <link rel="alternate    " href type>
    <link rel="feed alternate" href type>
    <link rel="alternate author" href type>

Here are a few examples that do *not* match:

    <link rel="alternatestuff" href type>
    <link rel="alternate" href>

Issue #277, opened 2017-12-27
Share -> Mail should use HTML


brentsimmons:
If possible, anyway. Otherwise it should use an attributed string.

vincode-io:
I didn't have any luck getting HTML passed through the Pasteboard.  It doesn't seem like any of the system share recipients will accept it including mail.  I didn't test it, but it might be possible to do a custom share service that uses Apple Events to talk to mail.  I'm not sure how desirable that would be with Mojave sandboxing these.

I did have success with NSAttributedString.  If you make the RTF UTI available in the Pasteboard the system share  recipients will take it.  The problem is that if you also offer up the URL UTI available they all will always select it over the RTF one.

I think the design question comes down to if we want to share the URL or the RTF via the Pasteboard?

brentsimmons:
Not going to do this one.

PR #460, opened 2018-09-14
Enabled RTF (NSAttributedString) pasteboard passing and removed the U…

vincode-io:
…RL passing.  Issue #277


vincode-io:
Per @brentsimmons we are going to stick with passing URL's and not do RTF since it impacts too many other sharing services besides mail.

PR #501, opened 2018-11-01
minor refactorings

adamnemecek:


danielpunkass:
I think the whitespace amendments make sense and might as well be accepted.

The decision to rewrite the #selector parameters in minimized form is more a question of taste. Because #selector's argument is interpreted by Swift and has to match against a known method name, it is able to infer from a standalone method name which actual selector to use. But to my mind it is better when the #selector argument actually conveys to anybody reading the code that there is an argument expected for the given method. Personally I would prefer to keep the somewhat noisy "(_:)" addendum on the #selector arguments.

danielpunkass:
Actually on the subject of the whitespace changes: I think it's probably a question of whether Brent cares about maintaining any whitespace standards and wants to establish some kind of standard going forward. If the editor is just going to add these kinds of whitespace in the future and it won't be enforced, there isn't really any point in taking a massive "correction" to whitespace.

vincode-io:
@adamnemecek thanks for taking the time to put together this pull request.  Unfortunately it is now out of date and has merge conflicts.

In the future if you want to implement some sweeping changes like this, please join the Slack channel and run them by Brent first.  It increases the chances that your pull request will be accepted.

Issue #134, opened 2017-11-07
Restore state at startup


brentsimmons:

brentsimmons:
Restore reader windows that were open, including window frames and splitview positions.

brentsimmons:
Backing out of multiple windows for 1.0. Instead, restore main window at startup, including splitview position.

tkrajacic:
Is there a specific reason not to use macOS' integrated mechanism for that? Sorry for the noise on closed issues.

[User Interface Preservation @ Apple docs](https://developer.apple.com/library/content/documentation/General/Conceptual/MOSXAppProgrammingGuide/CoreAppDesign/CoreAppDesign.html#//apple_ref/doc/uid/TP40010543-CH3-SW26)

brentsimmons:
I haven’t actually learned how to use it yet, so I did it old school. Ideally it would use the built-in mechanism, yes.

PR #549, opened 2019-02-15
Switch to standard state restoration mechanism

### LINKED ISSUE #547: State restoration should include selected feed/article
If I quit NetNewsWire with a specific feed/article selected, then the same feed and article should be selected when I relaunch the app.

danielpunkass:
I started looking at state restoration because I wanted to fix #547. I didn't quite figure out the best way to implement that but in the process of poking at it I revised the main window restoration approach to use standard state restoration. I noticed in #134 you mentioned wanting to use standard mechanism if possible so I offer this revision in that spirit. I don't think you need to include the manual window frame autosave name stuff either because I think it's automatically handled by AppKit's default state restoration.

### DIFF
--- a/NetNewsWire/AppDefaults.swift
+++ b/NetNewsWire/AppDefaults.swift
@@ -53,7 +53,6 @@ struct AppDefaults {
 		static let timelineSortDirection = "timelineSortDirection"
 		static let detailFontSize = "detailFontSize"
 		static let openInBrowserInBackground = "openInBrowserInBackground"
-		static let mainWindowWidths = "mainWindowWidths"
 		static let refreshInterval = "refreshInterval"
 
 		// Hidden prefs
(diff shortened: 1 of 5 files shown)


rex4539:
This commit broke saving the window state on quit.

Now on every restart the window is reset and appears really small.

danielpunkass:
@rex4539 Thanks for letting me know. Can you tell me anything more about your system? Is it a 10.14.3 macOS or something else?

Also can you verify that even after you change the window size back to something big, that quitting and reopening NetNewsWire it shows up small again? I wouldn't be surprised if existing saved state from before this commit is not restored.

rex4539:
I'm on 10.13.6

I can verify that changing anything on the window, quitting and relaunching, the app is reset to defaults.

danielpunkass:
@rex4539 Thanks. I'll take a closer look. I'm on 10.14 so I haven't tested the changes against 10.13.

danielpunkass:
I am able to reproduce the small window problem described by @rex4539, when I build and run the app on 10.13.

danielpunkass:
I opened ticket #550 to track the fix for this issue.

PR #571, opened 2019-03-03
Added ability to move between timeline and detail using arrows.  Issue #267

vincode-io:
This also enables the global shortcuts when in the detail view.


brentsimmons:
@vincode-io This PR changes the development team — would you resubmit it without that change? Thanks!

vincode-io:
Ok, there is now a commit in this pull request that undoes those team changes.

PR #576, opened 2019-03-04
Switch to ad hoc code signing, and disable the hardened runtime, for …

danielpunkass:
…Debug builds. This should make it easier for folks to clone the repository and get right to building/testing without configuring code signing details.

### DIFF
--- a/NetNewsWire.xcodeproj/project.pbxproj
+++ b/NetNewsWire.xcodeproj/project.pbxproj
@@ -2861,7 +2861,7 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = NO;
 				CODE_SIGN_ENTITLEMENTS = "Safari Extension/Subscribe_to_Feed.entitlements";
-				ENABLE_HARDENED_RUNTIME = YES;
+				CODE_SIGN_IDENTITY = "-";
 				INFOPLIST_FILE = "Safari Extension/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @executable_path/../../../../Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.ranchero.NetNewsWire-Evergreen.Subscribe-to-Feed";
@@ -3284,7 +3284,7 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				CODE_SIGN_ENTITLEMENTS = "NetNewsWire/NetNewsWire-MAS.entitlements";
-				ENABLE_HARDENED_RUNTIME = YES;
+				CODE_SIGN_IDENTITY = "-";
 				INFOPLIST_FILE = "NetNewsWire/Info-MAS.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks";
 				MACOSX_DEPLOYMENT_TARGET = 10.14;
@@ -3317,7 +3317,7 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = NO;
 				CODE_SIGN_ENTITLEMENTS = "Safari Extension/Subscribe_to_Feed.entitlements";
-				ENABLE_HARDENED_RUNTIME = YES;
+				CODE_SIGN_IDENTITY = "-";
 				INFOPLIST_FILE = "Safari Extension/Subscribe-to-Feed-MAS-Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @executable_path/../../../../Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = "com.ranchero.NetNewsWire-Evergreen.mas.Subscribe-to-Feed";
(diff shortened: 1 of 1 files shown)


jeff-h:
I guess a PR is hardly the right place for support, but my issue seems pretty related to this PR so I hope you don't mind. I've just cloned the repo (and submodules) and tried to build for macOS (non-MAS).

I noticed under Signing (both debug and release), the team is in red and says "Unknown Name (M8L2W...)" I changed these to my "personal team" but when I build I get these two errors:

* :-1: No account for team "M8L2WTLA8W". Add a new account in the Accounts preference pane or verify that your accounts have valid credentials. (in target 'Account')

* :-1: No signing certificate "Mac Development" found: No "Mac Development" signing certificate matching team ID "M8L2WTLA8W" with a private key was found. (in target 'Account')

I also noticed under capabilities that "Hardened Runtime" is turned on.

vincode-io:
@jeff-h, there are notes in the xcconfig/NetNewsWire_target.xcconfig file that will help you set up local team development for NetNewsWire.

meissnem:
@jeff-h Would you mind trying this patch? If it fixes your issue, I'll do a formal pull request.

[NNWproject.patch.txt](https://github.com/brentsimmons/NetNewsWire/files/3034609/NNWproject.patch.txt)

jeff-h:
@meissnem I tried the patch with no luck. I wouldn't rule out the usefulness of your patch based on my results though... for one thing, Xcode instantly spits out signing errors without even appearing to attempt to compile anything at all. It's like I changed nothing.

@vincode-io thanks for pointing that out! I did create the file and can confirm it's in the correct file path because the build scripts are picking up my team ID as entered in the file.

Would you mind telling me what this file should contain? I have tried every combination I can think of without success. It's currently like so:

```
CODE_SIGN_IDENTITY = Mac Developer
DEVELOPMENT_TEAM = 87TM2634T8
CODE_SIGN_STYLE = Automatic
PROVISIONING_PROFILE_SPECIFIER = com.example.appname
```

For `sss` I have tried both `Automatic` and `Manual`.

For `PROVISIONING_PROFILE_SPECIFIER =` I have tried three ways:

* omit the line entirely
* leave it in, but empty ie `CODE_SIGN_STYLE =`
* set it to `PROVISIONING_PROFILE_SPECIFIER = com.example.appname`
 
With the file set as above, I am getting five build errors, all similar to the following, for RSDatabase, Account, RSCore, RSTree, and RSWeb:

> error: RSDatabase has conflicting provisioning settings. RSDatabase is automatically signed, but provisioning profile com.example.appname has been manually specified. Set the provisioning profile value to "Automatic" in the build settings editor, or switch to manual signing in the project editor. (in target 'RSDatabase')


Sorry this is so painful. Very grateful for any suggestions.

vincode-io:
Here is my exact working xcconfig file.  You should only have to change the ```DEVELOPMENT_TEAM```.

```
CODE_SIGN_IDENTITY = Mac Developer
DEVELOPMENT_TEAM = SHJK2V3AJG
CODE_SIGN_STYLE = Automatic
PROVISIONING_PROFILE_SPECIFIER =
```

That, with a clean checkout of NetNewsWire should get you going.

meissnem:
@vincode-io Sorry I haven't been clear -- I'm using an xcconfig file exactly like yours. I ran into a problem with code signing because Brent's name is hardcoded in the shell script that signs Sparkle (?), but I see now that only affects Release builds and probably has nothing to do with @jeff-h 's situation.

jeff-h:
@vincode-io thanks for sharing your working file. Mine is now identical other than the team ID. I checked out a fresh copy of the repo and it still doesn't build. Could it be because I don't have a paid developer ID? There are five of the following pairs:

`error: No account for team "87TM2634T8". Add a new account in the Accounts preference pane or verify that your accounts have valid credentials. (in target 'RSDatabase')`

`error: No signing certificate "Mac Development" found: No "Mac Development" signing certificate matching team ID "87TM2634T8" with a private key was found. (in target 'RSDatabase')`

I also tried the shell script (`buildnnw`) although I assume it is for release builds? It doesn't appear to pick up my `SharedXcodeSettings/DeveloperSettings.xcconfig` so also fails.

jeff-h:
Wow, it built! I had to go through the five failing targets (NetNewsWire, Account, RSCore etc etc) and select my name from the Target > General > Signing > Team dropdown. Initially, each one looked like this:

Screen Shot 2019-04-04 at 12 59 25 PM

Of course, now the `*.pbxproj` in NetNewsWire and in all my submodules have changed and as such are showing in Git. I'm sure this is what the `DeveloperSettings.xcconfig` setup was intended to avoid. I'm more than happy to work with anyone who understands things better than I, if it would help resolve any issues with the build scripts. Otherwise, I guess the problem is somehow just unique to my setup.

Thanks all for your help, very much appreciated, especially on a project that quite clearly states it's not actively looking for PRs as yet.

brentsimmons:
I’d like to make it so you can do a Debug build without going through all that — but I have no idea *how*. It would be super-helpful, so future people won’t have to go through all this.

jeff-h:
Wow, insane :rage: I believe I have discovered why it refused to work for me. Probably everyone knows this except me heheh. It turns out there are MULTIPLE ten-character digit/uppercase strings associated with your certificate. I was using the one that shows when you hover over your Signing Certificate, as per `87TM2634T8` in this screenshot:

Screen Shot 2019-04-04 at 1 38 21 PM

But noooo. The correct one is obtained by going to the Keychain app, searching your Certificates for the "Mac Developer" (or "iPhone Developer) certificate. Double-click that, expand the details pane and find the "Organisational Unit". That will have a totally different ten-character code. 

Screen Shot 2019-04-04 at 1 44 37 PM

Using this one fixed everything instantly for me. The build script works beautifully as-is, and nothing in Xcode has to be edited.

Thanks again everyone for your help and for encouraging me to persist!

brentsimmons:
Wow! I would not have known to suggest this. Good detective work!

vincode-io:
I saw a project, [DevSwitch](https://github.com/aaronpearce/DevSwitch), the other day that seem to have this figured out.  They manage the different contributor teams the same way that the Brave browser does.  I did some analysis on it and the only thing I think I see that they do differently is this:

- The project file is configured for Automatic provisioning.
- The user specific xcconfig file only has the team in it.

I think their whole system of handling code signing is pretty slick.  We might want to emulate it.

https://github.com/aaronpearce/DevSwitch

PR #648, opened 2019-05-19
 Fixed a crash that occurs when trying to add a new feed with no active accounts

philviso:
#### Description

Fixed a crash that occurs when you have no active accounts, but try to add a new feed.

#### Details

The actual crash was occurring in `AddFeedWindowController.addFeed(_:)` as a result of force unwrapping the selected container. To fix the crash, I disabled the add feed toolbar item when there are no active accounts. Since the add feed toolbar item is now disabled when there are no active accounts, I also disabled the add folder toolbar item.

#### Question

At this point, the only toolbar item that's enabled when you have no active accounts is refresh all feeds. It doesn't hurt to leave it enabled, but it looks weird having *that* toolbar item enabled when all of the rest are disabled. Should we disable that toolbar item as well?

### DIFF
--- a/Mac/MainWindow/MainWindowController.swift
+++ b/Mac/MainWindow/MainWindowController.swift
@@ -227,6 +227,10 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 
 			return true
 		}
+		
+		if item.action == #selector(showAddFeedWindow(_:)) {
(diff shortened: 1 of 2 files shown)


brentsimmons:
Yes -- we should disable the refresh toolbar item (and menu command) when there are no active accounts.

PR #668, opened 2019-05-20
Improved size and layout of add feed / folder popover on iPad

### LINKED ISSUE #649: The Add Feed/Folder popover needs to be sized correctly on the iPad

philviso:
#### Description

Improved size and layout of add feed / folder popover on iPad. Also fixed a bunch of iOS issues I noticed along the way.

#### Details

* In addition to improving the sizing of the popover, I also cleaned up the margins and added support for dynamic type. Also, for both the add feed and folders UI, if the related picker only had 1 item to display, I hid the picker entirely.
* Disabled the add feed / folder button when there are no active accounts
* In the settings screen, I moved the "Add NetNewsWire News Feed" to the about section, and hid it when no active accounts.
* Fixed an issue on iOS where it always referred to local accounts as "On My iPhone", and standardized the "On My..." localized strings across the whole app.
* Updated a bunch of places that have text fields to dismiss the keyboard when return is tapped.
* Updated a few table views to allow allow certain cells to be selectable (e.g., the delete button in `DetailAccountViewController` is selectable, but the rest of the cells are that screen are not selectable).

Fixes #649

### DIFF
--- a/iOS/MasterFeed/MasterFeedViewController.swift
+++ b/iOS/MasterFeed/MasterFeedViewController.swift
@@ -49,7 +49,7 @@ class MasterFeedViewController: ProgressTableViewController, UndoableCommandRunn
 	}
 
 	override func viewWillAppear(_ animated: Bool) {
-		clearsSelectionOnViewWillAppear = splitViewController!.isCollapsed
+		clearsSelectionOnViewWillAppear = true
 		navigationController?.title = NSLocalizedString("Feeds", comment: "Feeds")
 		super.viewWillAppear(animated)
 	}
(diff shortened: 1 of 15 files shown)


iOS/MasterFeed/MasterFeedViewController.swift
```diff
@@ -48,7 +51,7 @@ class MasterFeedViewController: ProgressTableViewController, UndoableCommandRunn
 	}
 
 	override func viewWillAppear(_ animated: Bool) {
-		clearsSelectionOnViewWillAppear = splitViewController!.isCollapsed
+		clearsSelectionOnViewWillAppear = true
```
philviso: On iPad, the master feed cell is always left selected, because of how the master feed takes you to the timeline in the same view controller.

PR #675, opened 2019-05-24
Present add account view controllers modally + crash fix

### LINKED ISSUE #639: iOS: Action bar button use causes crash on iPad

### LINKED ISSUE #674: iOS Add Account UI should work like iOS Settings Add Account
Let's change the add account flow to work like the iOS settings Add Account flow with the modal dialog so that it is familiar.

philviso:
* Updated add account flows to present add account view controllers modally (Fixes #674)
* Fixed an exception that occurred when trying to present the share sheet on iPad, because it didn't have an anchoring view (Fixes #639)

### DIFF
--- a/iOS/Detail/DetailViewController.swift
+++ b/iOS/Detail/DetailViewController.swift
@@ -168,6 +168,8 @@ class DetailViewController: UIViewController {
 		}
 		let itemSource = ArticleActivityItemSource(url: url, subject: navState?.currentArticle?.title)
 		let activityViewController = UIActivityViewController(activityItems: [itemSource], applicationActivities: nil)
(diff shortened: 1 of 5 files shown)


iOS/Detail/DetailViewController.swift
```diff
@@ -168,6 +168,8 @@ class DetailViewController: UIViewController {
 		}
 		let itemSource = ArticleActivityItemSource(url: url, subject: navState?.currentArticle?.title)
 		let activityViewController = UIActivityViewController(activityItems: [itemSource], applicationActivities: nil)
+		activityViewController.popoverPresentationController?.barButtonItem = self.actionBarButtonItem
```
philviso: This is the crash fix. The rest of the PR is related to add account presentation stuff.

PR #715, opened 2019-05-28
Use 3 lines for timeline title + updated unread indicator layout

### LINKED ISSUE #690: iOS: Timeline: make left margin a little smaller
There’s more space between the left edge and the unread indicator than we need. We should probably use Mail’s metrics for this.

### LINKED ISSUE #694: iOS: Timeline: use three rows of text
Right now we have two. Let’s go with three and live with it a while and see if we prefer it.

philviso:
* Updated the timeline cell title to default to using 3 lines instead of 2
* Updated the timeline cell unread indicator left margin to match Mail.app's left margin
* Updated the timeline cell unread indicator size to match Mail.app

Fixes #690 
Fixes #694

### DIFF
--- a/iOS/MasterTimeline/Cell/MasterTimelineDefaultCellLayout.swift
+++ b/iOS/MasterTimeline/Cell/MasterTimelineDefaultCellLayout.swift
@@ -11,10 +11,10 @@ import RSCore
 
 struct MasterTimelineDefaultCellLayout: MasterTimelineCellLayout {
 
-	static let cellPadding = UIEdgeInsets(top: 8, left: 16, bottom: 8, right: 16)
+	static let cellPadding = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 16)
 	
-	static let unreadCircleMarginLeft = CGFloat(integerLiteral: 8)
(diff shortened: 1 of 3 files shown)


iOS/MasterTimeline/Cell/MasterTimelineDefaultCellLayout.swift
```diff
@@ -11,10 +11,10 @@ import RSCore
 
 struct MasterTimelineDefaultCellLayout: MasterTimelineCellLayout {
 
-	static let cellPadding = UIEdgeInsets(top: 8, left: 16, bottom: 8, right: 16)
+	static let cellPadding = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 16)
 	
-	static let unreadCircleMarginLeft = CGFloat(integerLiteral: 8)
-	static let unreadCircleDimension = CGFloat(integerLiteral: 8)
+	static let unreadCircleMarginLeft = CGFloat(integerLiteral: 0)
```
philviso: Technically this isn't needed anymore. I can get rid of it if people think it's just dead code. I left it around in case we want to tweak the value in the future.

PR #725, opened 2019-06-06
Various Minor Feedbin Fixes

tgt:
I've been looking into implementing FeedWrangler syncing in NetNewsWire since the app isn't useful for me without it. Whilst doing that I've come across a few bugs that are also issues with the current Feedbin syncing. Fixes for them are included in this pull request.

Hopefully the commits are either self-explanatory or explained by the commit messages, but please let me know if you'd like clarification or steps to reproduce the bugs.

### DIFF
--- a/Frameworks/Account/Account.swift
+++ b/Frameworks/Account/Account.swift
@@ -298,9 +298,7 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
 	}
 
 	public func refreshAll(completion: @escaping (Result<Void, Error>) -> Void) {
-		delegate.sendArticleStatus(for: self) { [unowned self] in
-			self.delegate.refreshAll(for: self, completion: completion)
-		}
+		self.delegate.refreshAll(for: self, completion: completion)
 	}
 
 	public func syncArticleStatus(completion: (() -> Void)? = nil) {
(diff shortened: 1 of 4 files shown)


Frameworks/Account/Account.swift
```diff
@@ -298,7 +298,9 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
 	}
 
 	public func refreshAll(completion: @escaping (Result<Void, Error>) -> Void) {
-		self.delegate.refreshAll(for: self, completion: completion)
+		delegate.sendArticleStatus(for: self) { [unowned self] in
+			self.delegate.refreshAll(for: self, completion: completion)
+		}
 	}
```
vincode-io: @tgt Could you please move the sendArticleStatus into the FeedbinAccountDelegate in the refreshAll func?  Right between the refreshArticles and the refreshArticleStatus is where it should have been put.  This will keep the meaning of refreshAll for the Account to mean the same thing as the refreshAll for the AccountDelegate.

tgt:
Will do.

I've just realised that "Fix updating local feed list when no remote feeds." was totally the wrong fix and totally breaks things so I've reverted it. There's a bug with deleting feeds on the server not being reflected in the client, but that wasn't the fix. I haven't managed to reliably reproduce the bug unfortunately.

Issue #719, opened 2019-06-01
No Article Menu item to mark article as unread


MarcusP-P:
When I go to the Article menu, on a read article, there is no menu item to mark an article as unread. Ideally Command-Shift-U would be a toggle to mark an article as read or unread.
![Screen Shot 2019-06-01 at 3 29 32 pm](https://user-images.githubusercontent.com/8262467/58744204-c22f3780-8482-11e9-9e87-09793ae8b04d.png)

PR #726, opened 2019-06-06
Toggles the menu item to mark an article as read or unread in the main menu

cparnot:
This should fix issue #719 

Because it's a menu item from the main menu, the initial "callback" for the menu is implemented in the app delegate, using the `NSMenuDelegate` (which is what is sanctioned by Apple in their docs) for the menu `Article`. The implementation in the app delegate calls into a new API in `MainWindowController` that is tasked with updating a "toggle read" menu item. The implementation in main window controller then passes that to the timeline view controller where the `Article` logic is implemented, and where it's the best place to actually do the work. This is just done in 4 lines using the existing business logic of the timeline view controller that knows wether selected articles can be marked as read.

The action for the 'Mark As Read/Unread' menu item has been changed to `toggleRead(_:)` since this is in fact the new behavior. Note that this action is already implemented to take care of the corresponding toolbar item in main window controller. And thus that's part of the responder chain for the menu item already. The implementation of `toggleRead(_:)` goes back down into the timeline view controller, which contains the appropriate business logic to take care of articles.

So this all seemed to me like the most consistent way to deal with the issue given what's already in place. The changes are minimal and anyone looking into how the `Mark as Read/Unread` menu item works should easily find its way.


cparnot:
Investigating more as there are other issues with the "Article" menu that might hint at a more general solution or a simpler fix.

PR #727, opened 2019-06-07
Fix issue with 'Mark as Read' menu item in the Article menu.

cparnot:
There were 2 issues: the title of the menu item did not toggle between 'Mark As Read' and 'Mark As Unread', and it was not invalidated where there is no article selected. All the code was already in place in `MainWindowController` as part of the interface item validation, but the menu item was using the wrong selector: `markRead` instead of `toggleRead`. Just changing the action in the menu item was needed to get the validation and menu item update happening correctly.


cparnot:
I forgot to mention the issue number in the commit: issue #719

RobFog:
It's probably a good idea amend the commit message to include the issue number.

PR #737, opened 2019-06-10
Adding CircleCI configuration for NetNewsWire (Mac only)

heckj:
This adds CircleCI configuration to setup NetNewsWire with CI. This really only applies for the Mac side of things right now, as the IOS side is requiring Xcode11 - and the latest XCode we can get on CircleCI is 10.2.1.

I've additionally tried to expand the CLI build with testing, but the tests are primarily osascript based, and the compilation for those tests is failing for me locally.

We *could* potentially add a similar CI configuration to each of the submodules and set up/extend testing there if that was desired.

In it's current configuration, this CI setup really only gives you a "it's compiled cleanly" (or not) indicator - not really anything about testing, coverage metrics, etc.

Current builds with this configuration can be seen under  https://circleci.com/gh/heckj/NetNewsWire, following the merge we'll get this set up under Brent's account linked to GH.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,58 @@
+# iOS CircleCI 2.0 configuration file
+#
+version: 2
+jobs:
+  build:
+
+    # Specify the Xcode version to use
+    macos:
+      xcode: "10.2.1"
+    # https://circleci.com/docs/2.0/configuration-reference/
+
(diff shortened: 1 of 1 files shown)


heckj:
@brentsimmons updated after @vincode-io's updates earlier today - so ready to merge. I'll be sending you instructions so we can get you set up with CircleCI for an OSS project shortly.

Issue #738, opened 2019-06-10
"Mark as Read" toolbar item label should be "Toggle Read"


Wevah:
(In the config sheet.)

…Since the behavior was changed back in 7c37ed5.

PR #739, opened 2019-06-10
Change label of "Mark Read" toolbar item to "Toggle Read"

Wevah:
For #738.


vincode-io:
@Wevah, could you file an issue for this and see if @brentsimmons agrees that this should be changed?  Changes like this that might be subjective are best done by creating an issue and waiting to see if it gets added to a release before doing a pull request.

Wevah:
Already have; sorry for not checking what the protocal was for PRs first.

PR #741, opened 2019-06-11
AppleScript: Make read/starred read/write

Wevah:
This patch makes read/starred status of articles read/write via AppleScript. Not sure `markArticles()` is the best way to make the UI update (since I don't know the codebase very well yet), but just setting the property on `ArticleStatus` obviously doesn't update the UI.


vincode-io:
Same as with the last request.  This is a best filed as a feature request and then see if it gets accepted into a release.  After that would be a good time to submit a pull request.

Wevah:
Sure.

PR #766, opened 2019-06-19
Google reader compatible syncing

jbeker:
Ticket: https://github.com/brentsimmons/NetNewsWire/issues/716

## NOTE 

This is not ready to merge, see "issues" below. I have hit the point where I could use some suggestions/help. 

## Overview

Implements support for syncing services that use the Google Reader API (specifically built against [FreshRSS](https://www.freshrss.org)). 

Includes support for:
* Syncing read/unread/starred content
* Subscribing to and unsubscribing from feeds
* Adding feeds to "folders" (tags)
* Removing tags from feeds
* Deleting unused tags
* Moving feeds to new folders

## Issues

* OPML import is not really functional. Unlike Feedbin, there is no API for importing OPML, so I was reduced to parsing the file and subscribing to each feed individually. This does not work in practice as the API response is too slow and things time out. Unfortunately I am at a loss as to how to do this better, so I could use suggestions/assistance
* Naming. "Google Reader Compatible API" is too long a name. I'd like suggestions for a better public facing description (this will be a bigger issue on iOS where screen space is at a premium)
* Icon. I didn't have an icon to represent this. Ideas?
* Removing tags (folders). There was a bug (https://github.com/FreshRSS/FreshRSS/issues/2411) in the shipping version of FreshRSS that prevents deleting tags. It appears to have been fixed (https://github.com/FreshRSS/FreshRSS/pull/2412). If you don't have a recent copy, you may need to update.
* There is also an oddity in that the API doesn't allow feeds at the top level, so they get dropped into the "Uncategorized" folder. This sometimes causes some oddities.

## Testing

You will need a FreshRSS installation to test this (or another Google API server, although I have not tested with any others). Ping me on the NetNewsWire Slack and I may be able to set you up with a test account.


vincode-io:
Regarding OPML, I suspect we will have to disable it for some sync services.  I don't think it is widely supported.

Let's just call this Reader API for now.  Maybe we should add a description under the name in the Add account area.

Let's just use a generic RSS icon for now.  I can add it after we merge.

Issue #772, opened 2019-06-20
iOS Suggestion: Use tag image instead of folder image?


stuartbreckenridge:
I amended the `masterFolderImage` to use SFSymbols `tag`.

```swift
static var masterFolderImage: RSImage = {
		let image = RSImage(systemName: "tag")!
		return image.maskWithColor(color: AppAssets.masterFolderColor.cgColor)!
}()
```

<img width="408" alt="Screenshot 2019-06-20 at 5 37 02 pm" src="https://user-images.githubusercontent.com/7046652/59838745-09be2a80-9382-11e9-87d6-a3ab882cc767.png">

stuartbreckenridge:
Closing as agreed to use folder image from SFSymbols.

PR #782, opened 2019-06-21
SF Symbols Expanded

### LINKED ISSUE #781: iOS Suggestion: Background app refresh should not happen when device is in low power mode

stuartbreckenridge:
Related to #772 - Updated more icons to use SF Symbols. Changed to using `withTint..` so that there is no aliasing.


philviso:
FYI, looks like background app refresh is disabled system wide whenever low power mode is enabled. We shouldn't need to do any extra handling.

More info right around the 3:15 mark in this session...

https://developer.apple.com/videos/play/wwdc2019/712/

stuartbreckenridge:
@philviso You are correct. I've rolled that part of the PR back. Thanks for pointing it out!

stuartbreckenridge:
@vincode-io d6ea5bd deletes the images that are not used.

PR #815, opened 2019-07-28
Cleaned up some .xcconfig files

olofhellman:
Moved a bunch of build settings out of the project.pbxproj file and into .xcconfig files.

As part of this change:

Changed the name of netnewswire_target.xcconfig to netnewswire_macapp_target.xcconfig Changed the name of netnewswire_ios_target.xcconfig to netnewswire_iosapp_target.xcconfig 

Created a separate .xcconfig file for the Safari extension

Added a run script build phase which will help keep buildSettings out of the project.pbxproj file. This script (VerifyNoBuildSettings) parses through the .pbxproj file and attempts to verify that no build settings are actually embedded there, and fails the build if it finds any.

Alphebetized a bunch of the settings in the existing xcconfig files

Added more information to the instructions for setting up a DeveloperSettings.xcconfig file, so that things should go more smoothly with regard to code signing if one needs to specify different CODE_SIGN_IDENTITY values for Mac and iOS builds

Added some helpful info to the names of the RunScript build phases


olofhellman:
So, I think the CI fix was to make sure to have 

    CODE_SIGN_IDENTITY[Config=Debug] = -

in the various xcconfig files.  Additionally, Xcode seems to have a bug where using the PRODUCT_NAME setting in an App Extension's xcconfig file doesn't actually set the Product Name, so we need to exempt PRODUCT_NAME from the restrictions on having BuildSettings in the pbxproj

PR #839, opened 2019-08-13
Break long words or URLs to prevent horizontal overflow

nostodnayr:


Wevah:
Note `word-wrap` is an alias for the standard's `overflow-wrap`. Shouldn't need to be changed, though.

Issue #857, opened 2019-08-15
Feature request: open in browser should work on multiple selection


brentsimmons:

hugolundin:
I have almost finished implementing this (pull request is coming soon), but need some guidance. 

Consider that the user selects multiple articles in the `TimelineView`. The first article selected will be marked as read – since it's shown in the `DetailView`. The consecutively selected ones are **not**, because its contents aren't shown in any view.

When opening these articles in the browser,  should they be marked as read or not? Is the general convention that opening an article in a browser considers it to have been read?

brentsimmons:
Reopening because this needs additional verification.

DavidMunch:
I can verify that it doesn't work per version 6.1.6 (6127). If you select more than one article, the menu option becomes greyed out, and the shortcut doesn't work as expected.

It would be very nice indeed to be able to select multiple article, and open them all at once, instead of having to do it manually for each article.

macOS 15.4.1.

PR #863, opened 2019-08-19
Support opening multiple selected articles in the browser at once

hugolundin:
This pull request implements #857. 

My approach was to look for all interactions open an article in the browser, and replace that with an implementation supporting multiple. Since a collection can contain a single item, this enabled the new functionality while preserving the existing. 

All these interactions start with the user selecting one or more articles in the `TimelineView`. Then one of the following things can be done:
- Press `Article` in the menu bar, and select `Open in Browser`.
- Press `⌘ + →`.
- Press enter.
- Right-click and select `Open in Browser` from the context menu.
- Press the globe icon in the right corner of the application.
- Double-click on *one* of the selected articles. I was uncertain whether this should open all selected items or only the one that is double-clicked. To make a decision I looked at how the built-in applications behave – they do the former (see for example Mail.app or Finder).

Before I consider this implementation to be done I need some advice about when to consider an article to be read. Today when an article is selected it is shown in the `DetailView` and therefore marked as read. When multiple articlesare selected, only the first one (that was shown in the `DetailView` before it says "Multiple selection") is marked as read. 

**Do we consider an article opened in the browser to be read?** Today's "Open in Browser" functionality does **not** care about read status at all – which leads to it to for example be possible to right click an article, open it in the browser and it will still be marked as unread in the app. 

Selecting and opening multiple articles in the browser makes this behaviour more prominent in more interactions, and I think it would be wise to make a decision on how it should be handled for future implementations involving this too.

### DIFF
--- a/Mac/MainWindow/MainWindowController.swift
+++ b/Mac/MainWindow/MainWindowController.swift
@@ -183,7 +183,7 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 	public func validateUserInterfaceItem(_ item: NSValidatedUserInterfaceItem) -> Bool {
 		
 		if item.action == #selector(openArticleInBrowser(_:)) {
-			return currentLink != nil
+			return currentLinks != nil
 		}
 		
 		if item.action == #selector(nextUnread(_:)) {
@@ -244,9 +244,11 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 	}
 
 	@IBAction func openArticleInBrowser(_ sender: Any?) {
-		if let link = currentLink {
-			Browser.open(link)
-		}		
+		if let links = currentLinks {
+			for link in links {
+				Browser.open(link)
+			}
+		}
 	}
 
 	@IBAction func openInBrowser(_ sender: Any?) {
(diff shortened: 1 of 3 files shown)


brentsimmons:
Thanks!

This needs two more things:

1. If the number of selected articles is above a certain number, then just ignore it entirely. Assume it’s an error. Let’s go with 500 for that number.

2. If the number of selected articles is above 20, then we need a confirmation sheet along the lines of: “Are you sure you want to open [number] pages in your browser?” [Cancel] [Open [number] pages]

hugolundin:
Will fix, both things seem really reasonable 👍 Thanks! 

How about marking them as read when opened in the browser?

vincode-io:
@hugolundin Are you still working on @brentsimmons changes?  I'm just trying to figure out if this is a zombie pull request or if you are still interested in it.

hugolundin:
Yes, sorry. I started my first year at a computer engineering program 4 weeks ago and haven't had time to complete it. Looking to do it soon, if it is OK? :)

vincode-io:
Absolutely!  Take your time.  I just don't like leaving dead pull requests laying around, so I asked.

hugolundin:
I haven’t and probably won’t have time to work in this for a while. Happy to let someone else work in it if they want to.

vincode-io:
Thanks for letting us know!  I'll close this pull request.

PR #912, opened 2019-08-30
#914 - Explicity identify the addable account types.

kielgillard:
Allowing for a contingent list given factors such as macOS API availability.

The specific use case I have in mind is offering accounts based on the availablility of [AuthenticationServices for web authentication](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession) in Catalina.


kielgillard:
Implements #914

Issue #921, opened 2019-08-31
"Open in Browser" is enabled when no article is selected


DamianLettie:
Click on an empty spot in the article list. The `Article -> Open in Browser` menu item is enabled, even though no article is selected. The menu item does nothing if clicked.

Tested NetNewsWire 5.0 (2609) on macOS 10.14.6.

PR #929, opened 2019-08-31
Only enable "Open in Browser" when a single article is selected.

correia:
Fixes bug #921.


correia:
Closing to resubmit from correct branch.

PR #945, opened 2019-09-02
Add optional separators in the timeline view.

correia:
The timeline now optionally includes "Mail style" separators (behind a
hidden default — "CorreiaSeparators" — which defaults to NO).

I find the separators make reading the timeline considerably more comfortable.

Before:
before

After:
after


gferreroferri:
How can I change this default from the Terminal? Where is this stored?

gferreroferri:
Found it: 
defaults write com.ranchero.NetNewsWire-Evergreen CorreiaSeparators -bool YES

:)

PR #1016, opened 2019-09-13
Add setting to group articles by feed in timeline

philviso:
## Description

Added setting to group articles by feed in the timeline for both Mac and iOS. For Mac, the toggle is in the "View" menu. For iOS, the toggle is in the settings screen in the "Timeline" section.

## Details

The coding guidelines for NNW say to use *generics* sparingly, so there's one part of this PR that could be slightly controversial. I initially implemented the sorting logic right in `ArticlesArray`, but when I went to write unit tests for my sorting logic I realized that `ArticlesArray` isn't easily testable (It depends on the `AccountManager` singleton).

To make the code more testable, I broke the sorting logic out into an `ArticesSorter` class and added a `SortableArticle` protocol. `ArticleSorter.sortByDate(_:groupByFeed)` accepts an array of `SortableArticle` objects and uses generics to return the _actual_ type of `SortableArticle` (in our case, it ends up returning `[Article]`.

----

One *performance* implication I wanted to callout - `Article` has a `feed` calculated property that I needed to access in order to group by feed name. That property is calculated via multiple lookups - first looking up the account via `accountID` and then the feed in that account via `feedID`. That means these lookups are repeatedly happening during sorting.

Using this large data set (https://github.com/kilimchoi/engineering-blogs/blob/master/engineering_blogs.opml) with 16k unread articles, I didn't notice any performance difference on the Mac. That could be different on older iOS devices though. I don't have an old iOS devices to try it out with.

I didn't want to make major changes to this aspect of how `Account` works unless we think it actually matters. It's redundant to have to do the lookups over and over, but they are dictionary lookups so it should be fast.


Shared/Timeline/ArticleSorter.swift
```diff
@@ -0,0 +1,61 @@
...
+		
+	private static func sortedByFeedName<T: SortableArticle>(articles: [T],
+															 sortByDateDirection: ComparisonResult) -> [T] {
+		// Group articles by "feed-feedID" - feed ID is used to differentiate between
+		// two feeds that have the same name
+		let groupedArticles = Dictionary(grouping: articles) { "\($0.sortableName.lowercased())-\($0.sortableFeedID)" }
+		return groupedArticles
+			.sorted { $0.key < $1.key }
+			.flatMap { (tuple) -> [T] in
+				let (_, articles) = tuple
```
philviso: Never seen this swift error before, but it wouldn't let me define the (key, articles) tuple directly in the `flatMap` signature. It made me use a single `tuple` argument, and then restructure it here.

Tests/NetNewsWireTests/ArticleSorterTests.swift
```diff
@@ -0,0 +1,240 @@
...
+//  Copyright © 2019 Ranchero Software. All rights reserved.
+//
+
+import Articles
+import Foundation
+import XCTest
+
+@testable import NetNewsWire
+
+class ArticleSorterTests: XCTestCase {
```
philviso: For some reason, from this point down in the PR all of the indentation looks weird. I'm using whatever the defaults are in Xcode for indentation settings, so I'm not sure why GitHub is showing these huge indentations.

iOS/Settings/SettingsView.swift
```diff
@@ -54,7 +54,10 @@ struct SettingsView : View {
 	func buildTimelineSection() -> some View {
 		Section(header: Text("TIMELINE")) {
 			Toggle(isOn: $viewModel.sortOldestToNewest) {
-				Text("Sort Oldest to Newest")
+				Text("Sort Newest to Oldest")
```
philviso: @vincode-io Here's the copy fix for iOS settings that we talked about.

Issue #1006, opened 2019-09-11
Feature Request: Use Google search as a feed 'registry'


igor-makarov:
I was listening to the Gruber show and I heard it being said that there is no definitive 'RSS feed registry', similar to the iTunes Podcast Registry.  
I'm here to say that there is one, and it's Google.

More often than not, looking for a website by name on Google places it in the top 1 or 2 results.

While there's no direct Google Search API, there's a pretty easy way to extract the links from it.  
I've created this simple API in 15 minutes using off-the-shelf parts and published it as a [Netlify function](https://demo-google-function.netlify.com). Keep in mind that it only accepts a few specific query strings to prevent bot abuse.

Alternatively, this could be done on the client, and here's a small bashpipe that does it:
```bash
curl -H 'User-Agent: Mozilla/5.0 (Windows NT 6.1; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.85 Safari/537.36' \
  -s https://www.google.com/search?q=stratechery | \
  grep -Eoi '<a [^>]+>' |  \
  grep -Eo 'href="http[^\"]+"' | \
  grep -v 'google.'
```
<br/>

### TL;DR
It's relatively easy to use Google search to find the website that the user really wants, thus giving a similar experience to podcast search. I think NetNewsWire users would love this.

DamianLettie:
Be careful of [Google's Terms of Service](https://policies.google.com/terms). e.g.:
> don’t [...] try to access [our services] using a method other than the interface and the instructions that we provide

igor-makarov:
That's an interesting concern. One could argue that accessing Google Search by making a direct HTTP request to `https://www.google.com/search?q=stratechery` would be accessing it through the interface they provide, but I don't think it's up to me to argue this 🙃

I would, however, say, that overall usage of Google by the suggested feature would be microscopic in scale and will definitely not amount to abuse. It will only occur once per adding a feed action.

This leads me to believe that Google will have no interest in blocking this access.

PR #1022, opened 2019-09-15
[DNM] Use Google Search for finding feeds

igor-makarov:
This provides #1006 with a semi-implementation suggestion.

Allow me to preface with saying that this isn't a finished feature by any stretch. I'm posting this mostly to get the conversation started. If this isn't the best place to share my code, please let me know.

Furthermore, it doesn't actually work yet but I did manage to get it halfway. If you go to `Add Feed > Search for a Feed`, you can type queries and see which URLs come back. It's even debounced to reduce request rate.

The way it works is that it utilizes the existing framework for link scraping and scrapes a google search result page.

Once I've reached the website URLs phase, I've tried to figure out how to display them in a list with their actual feed URLs and favicons but this seems a fair amount more work so I wanted to know if this might actually be of interest.

Thank you for your time!

### DIFF
--- a/Frameworks/Account/FeedFinder/FeedFinder.swift
+++ b/Frameworks/Account/FeedFinder/FeedFinder.swift
@@ -11,9 +11,10 @@ import RSParser
 import RSWeb
 import RSCore
 
-class FeedFinder {
+public class FeedFinder {
 	
 	static func find(url: URL, completion: @escaping (Result<Set<FeedSpecifier>, Error>) -> Void) {
+
 		downloadUsingCache(url) { (data, response, error) in
 			if response?.forcedStatusCode == 404 {
 				completion(.failure(AccountError.createErrorNotFound))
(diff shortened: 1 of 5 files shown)


zackschuster:
worst case scenario: google hates rss & will find some way to use this to destroy or slow down the project

best case scenario: google doesn't inadvertently break this like it does everything else & simple search queries may be performed in peace

as google is already in many cases a walking worst case scenario, i would caution against this feature

brentsimmons:
My suggestion: this would be great for a separate app that’s focused on feed discovery.

An example of that kind of app is Feed Compass, which is free and on the Mac App Store: https://apps.apple.com/us/app/feed-compass/id1458648487?mt=12

vincode-io:
The NetNewsWire Slack has a channel that we discuss feed discovery in sometimes.  Check out #openweb if you want to bounce ideas around or show off some code.

PR #1026, opened 2019-09-17
Enable inspector

### LINKED ISSUE #1025: Add "Enable Web Inspector" menu item to the Debug menu
Tracking issue. (Currently just enables the "Inspect Element" contextual menu item in the WKWebView.)

I'm almost done with this, but I have one question: As it's a hidden API on `WKPreferences`, should I use `setValue(_,forKey:)`, or add a bridging header?

(Current WIP at https://github.com/Wevah/NetNewsWire/tree/enable-inspector.)

Wevah:
Fixes #1025.

Questions:

I'm setting the menu item state in `validateUserInterfaceItem()`. The current code sets the state for the sort order menu items manually. Should I change the implementation to follow that style?

A custom notification is currently sent when the setting is changed. Should it monitor `UserDefaults.didChangeNotification` instead?

### DIFF
--- a/Mac/AppDefaults.swift
+++ b/Mac/AppDefaults.swift
@@ -36,6 +36,8 @@ struct AppDefaults {
 		static let timelineShowsSeparators = "CorreiaSeparators"
 		static let showTitleOnMainWindow = "KafasisTitleMode"
 		static let hideDockUnreadCount = "JustinMillerHideDockUnreadCount"
+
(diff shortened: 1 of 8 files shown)


vincode-io:
This looks ok to me, so  I'm going to merge it.  If Brent wants the notification mechanism to change we can do it in a separate pull request.

PR #1027, opened 2019-09-17
Add a Feedly account with user consent

kielgillard:
The Feedly integration in general is a work in progress. Adding issues for the rough edges of this PR.

### DIFF
--- a/Frameworks/Account/Account.swift
+++ b/Frameworks/Account/Account.swift
@@ -227,6 +227,8 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
 			self.delegate = FeedbinAccountDelegate(dataFolder: dataFolder, transport: transport)
 		case .freshRSS:
 			self.delegate = ReaderAPIAccountDelegate(dataFolder: dataFolder, transport: transport)
+		case .feedly:
+			self.delegate = FeedlyAccountDelegate(dataFolder: dataFolder, transport: transport)
 		default:
 			fatalError("Only Local and Feedbin accounts are supported")
 		}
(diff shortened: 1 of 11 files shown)


NetNewsWire.xcodeproj/project.pbxproj
```diff
@@ -2194,16 +2200,16 @@
 				TargetAttributes = {
 					513C5CE5232571C2003D4054 = {
 						CreatedOnToolsVersion = 11.0;
-						DevelopmentTeam = SHJK2V3AJG;
+						DevelopmentTeam = 5G4WXPEDG7;
```
kielgillard: Revert this change before merging.

Mac/Preferences/Accounts/AccountsAddViewController.swift
```diff
@@ -15,7 +15,7 @@ class AccountsAddViewController: NSViewController {
 	
 	private var accountsAddWindowController: NSWindowController?
 	
-	private let addableAccountTypes: [AccountType] = [.onMyMac, .feedbin, .freshRSS]
+	private let addableAccountTypes: [AccountType] = [.onMyMac, .feedbin, .freshRSS, .feedly]
```
kielgillard: Revert this change before merging, too.

Frameworks/Account/Account.swift
```diff
@@ -307,6 +309,28 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
...
 		}
 	}
+	
+	public static func oauthAuthorizationCodeGrantRequest(for type: AccountType, client: OAuthAuthorizationClient) -> URLRequest {
+		switch type {
+		case .feedly:
+			return FeedlyAccountDelegate.oauthAuthorizationCodeGrantRequest(for: client)
+		default:
+			fatalError("\(type) does not support OAuth authorization code granting.")
+		}
```
vincode-io: We don't usually use fatalError and use assertionFailure instead.

kielgillard: For release style builds the assertion would be stripped and I'd be required to return a value for these cases. Returning an "empty" URL request seemed like a hack so I figured it was reasonable to fatalError.

brentsimmons: If I read this right: in this case it’s a serious programming error and not something that would normally happen, even in an edge case on a bad day. So I’m okay with a fatalError (or precondition check) for this.

kielgillard: That's how I understand the situation, too. The programmer is trying access behaviours of certain objects that by definition they are not capable of. I expect the most basic verification of this functionality during development will have revealed this to prevent it occuring in the wild.

NetNewsWire.xcodeproj/project.pbxproj
kielgillard: Reverted in fc5b003ab0523160bdc56e4a5ead8c313348e845.

Mac/Preferences/Accounts/AccountsAddViewController.swift
kielgillard: After talking with @vincode-io, it seems there is no good reason to revert this change.

Issue #697, opened 2019-05-25
iOS: Detail: Use JavaScript to set HTML


brentsimmons:
The issue is performance when going to next/previous article or next unread. Ben and Daniel have both mentioned using JavaScript to set the various bits of HTML — that that’s faster than loading a new HTML page.

We should try it. (And, probably, later use this on Mac too.)

benubois:
I tried hacking this in to an old build of Evergreen. It definitely was faster _except_ when there were images in the article. I think the sheer amount of base64 encoded image data made it slower. 

If you did do this, I think locally cached images would have to be referenced as files, rather than inserted into the document as base64 strings.

brentsimmons:
Sounds like this might be more work than is warranted for the 5.0 release. Let’s hold it for after 5.0.

PR #1033, opened 2019-09-19
Render with evaluateJavaScript.

benubois:
Replaced `loadHTMLString` based rendering for improved performance. This makes rendering articles have no visible lag, which currently happens in posts with external assets like images.

[Before](https://shared-files-and-data.s3-us-west-1.amazonaws.com/before.mov)
[After](https://shared-files-and-data.s3-us-west-1.amazonaws.com/after.mov)

The css is currently duplicated into the new wrapper template. It could move there or stay in the external `styleSheet.css` (maybe that's better for themes?).

There still might be an issue with articles that contain many large base64 encoded images. One workaround would be to fallback to `loadHTMLString` if the article is past a certain size. Would need to test to see where that limit is.

### DIFF
--- a/Mac/MainWindow/Detail/DetailWebViewController.swift
+++ b/Mac/MainWindow/Detail/DetailWebViewController.swift
@@ -100,7 +100,13 @@ final class DetailWebViewController: NSViewController, WKUIDelegate {
 
 		NotificationCenter.default.addObserver(self, selector: #selector(webInspectorEnabledDidChange(_:)), name: .WebInspectorEnabledDidChange, object: nil)
 
-		reloadHTML()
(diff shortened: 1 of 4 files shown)


benubois:
This would fix #697. It also fixes a new flash-of-styled content [introduced when stripping styles `onload`](https://github.com/brentsimmons/NetNewsWire/commit/8a8dcde3876e3e6f508f47bbd03a891c98635da1#diff-e71ab10edabb871175ddcc48f1048fe7).

vincode-io:
Ben, this is awesome.  Page loads are notably faster.

We have separate CSS stylesheets for iOS and the Mac to deal with different margins, fonts, and colors.  Let's keep the stylesheet separate from page.html to keep that working.

PR #1035, opened 2019-09-19
Disable Web Inspector for App Store builds

Wevah:
- Hide the "Enable Web Inspector" menu item.
- #if-out notifications/defaults/etc.
- Make toggleWebInspectorEnabled(_) a no-op.


Wevah:
Notes:

I didn't completely `#if`-out toggleWebInspectorEnabled because I'd have to rewire the action to the first responder to avoid showing a warning, and that felt weird. It's easy enough to switch to that if it's preferred, though.

PR #1036, opened 2019-09-19
Made iFrames responsive. 

### LINKED ISSUE #1034: Youtube embedded videos are very small
Youtube videos are rendering very small on the desktop app:

Screen Shot 2019-09-19 at 10 45 47

I tried fiddling with settings, but could not find anything to change that. And if I recall right, this was not happening a few days ago.

benubois:
This fixes Fixes #1034.

Screen Shot 2019-09-19 at 10 45 26 AM

### DIFF
--- a/Shared/Article Rendering/ArticleRenderer.swift
+++ b/Shared/Article Rendering/ArticleRenderer.swift
@@ -328,8 +328,6 @@ private extension ArticleRenderer {
 		return dateFormatter.string(from: date)
 	}
 
-	#if os(macOS)
-	
 	func renderHTML(withBody body: String) -> String {
 
 		var s = "<!DOCTYPE html><html><head>\n\n"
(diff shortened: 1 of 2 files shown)


Shared/Article Rendering/ArticleRenderer.swift
```diff
@@ -375,40 +388,6 @@ private extension ArticleRenderer {
 		return s
 	}
 	
-	#else
-	
-	func renderHTML(withBody body: String) -> String {
```
benubois: Is it ok to delete this ios block? As far as I could tell the only difference is the lack of `mouseenter`/`mouseleave` listeners. Should be ok to include these. iOS does not support these events so they should be ignored.

vincode-io: Yeah, delete that.

vincode-io:
Since this is a 5.0.3a issue, could you submit the pull request to the mac-release branch please?

PR #1039, opened 2019-09-22
Remove enableWebInspectorMenuItem outlet

Wevah:
Fixes a console warning about the removed property.


Wevah:
I forgot to remove an outlet to the Enable Web Inspector menu item from the storyboard after learning that the Debug menu was removed completely in release builds.

PR #1073, opened 2019-09-26
Support Daring Fireball style footnote markup (#544)

brehaut:
Original implementation was only aware of multimarkdown style footnote markup.

This commit includes support for DaringFireball style footnotes

### DIFF
--- a/Mac/MainWindow/Detail/styleSheet.css
+++ b/Mac/MainWindow/Detail/styleSheet.css
@@ -225,10 +225,19 @@ img[src*="share-buttons"] {
 	margin-bottom: 0;
 }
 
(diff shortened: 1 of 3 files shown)


vincode-io:
@brehaut This is merged, but I am still having problems with that one Daring Fireball article.

brehaut:
@vincode-io Thanks, I'm a bit confused about this one. I'll have a deeper look this weekend. Is it only that one DF article that is being trouble, or all of them?

vincode-io:
I haven't found any other articles that have footnotes yet.  Could you share a feed that has them so that I can verify against it?

Issue #1015, opened 2019-09-12
Feature Request: Swipe to toggle unread status ala Mail.app [macOS]


jjatie:
Swipe to mark as read/unread just like in Mail.app (should be same direction too for muscle memory).

Wevah:
I have working code for this if it's decided it should be added. Currently the actions are just text; images à la Mail would be good.

vincode-io:
I'm assigning this to 5.1 since we have working code and are going to try to get design assets for it.

vincode-io:
This will be part of the 5.1 release.

PR #1091, opened 2019-10-02
Swipe to mark

Wevah:
Patch for #1015. Still need graphic assets, which could probably be pulled from SF Symbols.

### DIFF
--- a/Mac/MainWindow/Timeline/TimelineViewController.swift
+++ b/Mac/MainWindow/Timeline/TimelineViewController.swift
@@ -746,6 +746,32 @@ extension TimelineViewController: NSTableViewDelegate {
 		cell.objectValue = nil
 		cell.cellData = TimelineCellData()
 	}
+
+	private func toggleArticleRead(_ article: Article) {
+		guard let undoManager = undoManager, let markUnreadCommand = MarkStatusCommand(initialArticles: [article], markingRead: !article.status.read, undoManager: undoManager) else {
+			return
+		}
+		self.runCommand(markUnreadCommand)
+	}
+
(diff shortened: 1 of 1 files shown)


Mac/MainWindow/Timeline/TimelineViewController.swift
```diff
@@ -746,6 +747,50 @@ extension TimelineViewController: NSTableViewDelegate {
...
+		}
+		self.runCommand(markUnreadCommand)
+	}
+
+	private func toggleArticleStarred(_ article: Article) {
+		guard let undoManager = undoManager, let markUnreadCommand = MarkStatusCommand(initialArticles: [article], markingStarred: !article.status.starred, undoManager: undoManager) else {
+			return
+		}
+		self.runCommand(markUnreadCommand)
+	}
```
Wevah: These could maybe be combined into a `toggleStatus()` function or somesuch, but I left them separate as that's how it's done in other code.

Mac/MainWindow/Timeline/TimelineViewController.swift
```diff
@@ -746,6 +747,50 @@ extension TimelineViewController: NSTableViewDelegate {
...
+				return [action]
+
+			case .trailing:
+				let title = article.status.starred ? NSLocalizedString("Mark Unstarred", comment: "mark unstarred") : NSLocalizedString("Mark Starred", comment: "mark starred")
+				let action = NSTableViewRowAction(style: .regular, title: title) { (action, row) in
+					self.toggleArticleStarred(article);
+					tableView.rowActionsVisible = false
+				}
+				return [action]
+
```
Wevah: Some redundant code here; could combine into a single call point if necessary.

PR #1093, opened 2019-10-03
Optimize PNG’s

majouji:
Ran PNG images through ImageOptim to reduce file sizes without compression.

Screen Shot 2019-10-02 at 10 52 42 PM


brentsimmons:
I just want to be sure — they are absolutely, 100% pixel-for-pixel the exact same after running through ImageOptim?

majouji:
> I just want to be sure — they are absolutely, 100% pixel-for-pixel the exact same after running through ImageOptim?

Yep! No compression, transform of any kind, stripped metadata, or changed color profiles. The exact same, just better :)

PR #1104, opened 2019-10-05
iOS: Added a setting to suppress the mark all as read alert

philviso:
* Added a toggle to settings to suppress the alert that displays before marking all articles as read. The default value matches the current behavior, so the alert still displays every time you tap mark all as read.
* There were two different mark all as read alerts that were identical except for a small copy difference, so I unified them.


vincode-io:
We took a different approach.

PR #1112, opened 2019-10-08
iOS: Only show mark all as read alert one time

philviso:
* Added some logic to only show the mark all as read alert the first time that any mark all as read button is tapped.
* Updated the alert copy to include how to undo mark all as read in the future
* Unified both of the mark all as read alerts into one class

#### Note
* Suppressing the alert is tied to actually tapping the "Mark All Read" action in the alert. Until the user actually _does_ a mark all as read, the alert will continue to appear.

### DIFF
--- a/iOS/MasterFeed/MarkArticlesReadAlertController.swift
+++ b/iOS/MasterFeed/MarkArticlesReadAlertController.swift
@@ -1,0 +1,44 @@
+//
+//  MarkArticlesReadAlertControllerr.swift
+//  NetNewsWire
+//
+//  Created by Phil Viso on 9/29/19.
+//  Copyright © 2019 Ranchero Software. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+struct MarkArticlesReadAlertController {
+		
+	static func allArticlesAlert(handler: @escaping (UIAlertAction) -> Void) -> UIAlertController {
+		let message = NSLocalizedString("Mark all articles in all accounts as read?",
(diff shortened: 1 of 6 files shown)


iOS/MasterFeed/MarkArticlesReadAlertController.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Copyright © 2019 Ranchero Software. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+struct MarkArticlesReadAlertController {
+		
+	static func allArticlesAlert(handler: @escaping (UIAlertAction) -> Void) -> UIAlertController {
+		let message = NSLocalizedString("Mark all articles in all accounts as read? You can undo this action with a three finger swipe to the left.",
```
philviso: Open to ideas about how to make the message also convey that NNW supports undo on most actions in the app. The message is already pretty long, so I was trying to keep it brief.

vincode-io: I'm concerned that the information about the undo is in the second sentence and the button still uses the Mark All Read.  I think people will just hit read that button, hit it, and never read the verbiage.

Maybe we should just have a dialog that says, "You can undo this and other actions with a three finger swipe to the left." with a "Got It" button.  My hope is that the "Got It" will give them some pause and make them read the verbaige.

philviso:
@vincode-io Updated

Issue #1080, opened 2019-09-27
Feedly: improve errors handling when creating or updating collections


kielgillard:
When a request to create or update a collection/folder fails because the name has invalid characters, uses the reserved names or whatever, we should show a user friendly error rather than nerdy ones like “400 bad request”.

Issue #1046, opened 2019-09-22
Feedly: Remove a feed


kielgillard:
To consider this done:

When the user removes a feed in a folder or at the top level in NNW, the feed should be removed from the collection or Uncategorized category in Feedly.

When the user removes a feed in a collection in Feedly, the feed should be removed from the corresponding folder or the top-level in NNW.

kielgillard:
Implemented. Please file specific issues if something doesn't work as expected.

Issue #1045, opened 2019-09-22
Feedly: Create, update or move a Feed


kielgillard:
Implement basic feed grouping.

To consider this done:
When the user adds a feed with a specific name in NNW to a Folder, the feed should appear with the same name in the corresponding collection in Feedly.

When the user adds a feed with a specific name in NNW as a top level, the feed should appear with the same name in the Uncategorized collection in Feedly.

When the user adds a feed in Feedly, it appears in the corresponding Folder or top level in NNW.

When the user renames an existing feed in NNW, the feed should use the new name in Feedly.

When the user renames a feed in Feedly, it appears renamed in NNW.

When the user moves a Feed from one Folder to another in NNW, the feed should be removed from the source collection and added to the destination collection in Feedly.

When the user moves a top level feed into a Folder in NNW, the feed should be removed from the Uncategorized collection to the destination collection in Feedly.

(To be confirmed): When the user moves a feed into the Uncategorized collection in Feedly, the feed should appear at the top level of the side bar in NNW.

kielgillard:
Depends upon #1044.
See https://developer.feedly.com/v3/collections/#create-or-update-a-personal-collection

kielgillard:
Implemented. Please file specific issues if something doesn't work as expected.

PR #1120, opened 2019-10-09
Implements creating, updating, moving and removing feeds.

kielgillard:
I'm taking a "get it working, then refactor" attitude. Now that it's working, I can see specific ways how things could be refactored for testability and reusability.


kielgillard:
Implements #1045 #1046

kielgillard:
Some work towards #1080

Issue #1041, opened 2019-09-22
Feedly: Send and receive starred status 


kielgillard:
Implement the send and refresh article statuses for the starred key. 

To consider this done:
When the user stars an article in NNW, the article should appear in the Read Later section on Feedly (refreshing the browser could be required).

When the user stars an article in Feedly, the article should appear in the Starred section in NetNewsWire after refresh or a period of time (?).

The use cases for unstarring an article are predictably similar. 

https://developer.feedly.com/v3/markers/#mark-one-or-multiple-articles-as-saved
https://developer.feedly.com/v3/tags/

RobFog:
This should be fixed with #1135, correct?

PR #1135, opened 2019-10-15
Send and receive starred articles…

kielgillard:
…refactoring status syncing code into operations for better code reuse.


kielgillard:
Implements #1041.

PR #1145, opened 2019-10-19
Disable "Add Folder" button when the text field is empty

Wevah:


brentsimmons:
Reminder: let’s merge this after 5.0.3 ships. It should go into 5.0.4.

PR #1148, opened 2019-10-21
Use an accessory view instead of an intermediate sheet for OPML export

### LINKED ISSUE #1142: Make OPML export one step — use accessory view

Wevah:
Fixes #1142.


Wevah:
Maybe “delegate” should be renamed; not sure.

PR #1152, opened 2019-10-22
MAS and sandboxing support

danielpunkass:
There are probably a few rough edges but it would be useful to incorporate this to start testing sandboxing support, and to generate an official build from which we can test updating via Sparkle to another official build.


danielpunkass:
Sorry for the massive number of files - it's because I deleted the Sparkle binary in the process of switching to a submodule. Let me know if any questions!

Issue #1198, opened 2019-10-27
Error sending 5.1d1 for notarization


brentsimmons:
Looks like there is some configuration still to do before we can get this notarized.

Or may it’s an issue with getting the right Sparkle branch? I’m not sure.

![Screen Shot 2019-10-26 at 8 53 41 PM](https://user-images.githubusercontent.com/1297121/67629445-d0fff780-f832-11e9-8a7b-2df0c4cea3a4.png)

PR #1201, opened 2019-10-27
Fix Notarization NOTE: ADDS FILE DELETION to build phase

danielpunkass:
NOTE: Please review the added shell script lines to sanity-check the file deletion logic. I think it's very safe because of the use of "find" and targeting it to a specific, quoted build path that should limit file deletion only to the Sparkle.framework path of the build product.

Add commands to the Sparkle build phase script to remove Autoupdater and Updater.app binaries. These are not needed by NetNewsWire since we use the XPC services to provide the same functionality.


danielpunkass:
This will hopefully fix #1198

PR #1247, opened 2019-11-04
Rearrange date label to right side of article list.

sbarex:
screenshot


brentsimmons:
We’re not taking pull requests for things we haven’t decided to do. If there’s particular work you’d like to do, please join the Slack group and talk it over on the #work channel. Thanks!

Issue #1029, opened 2019-09-18
Feedly: authenticate using ASWebAuthenticationSession?


kielgillard:
See if we can use the `ASWebAuthenticationSession` API introduced in Catalina/iOS 13 to simplify the experience of adding a Feedly account to NetNewsWire (NNW). As Apple's [documentation indicates](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession):
> If the user has already logged into the web service in Safari or other apps using ASWebAuthenticationSession, it’s possible to share the existing login information.

At the time of writing, `AccountsFeedlyWebWindowController` uses a `WKWebView` to load a `URLRequest` for the resource into which the user gives Feedly consent to grant NNW access tokens. Investigate passing the request's URL [into the initaliser](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/2990952-init), a callback scheme ([Feedly has rules](https://developer.feedly.com/v3/auth/#authenticating-a-user-and-obtaining-an-auth-code) for the callback URI) and in the completion handler, treat the URL like the web view's `WKNavigationDelegate`.

PR #1278, opened 2019-11-08
Use ASWebAuthenticationSession to obtain OAuth grant for Feedly production and sandbox.

kielgillard:
EDIT: This is now outdated. Please [see the final description](https://github.com/brentsimmons/NetNewsWire/pull/1278#issuecomment-552087841).

This PR:
- Allows NNW to request an OAuth grant using `ASWebAuthenticationSession`. These changes are added to the NNW, Mac App Store and iOS targets.
- Substitutes the production ("feedly cloud") OAuth client identifier and secret in a Run Script build phase. In my own testing, I added these values to `SharedXcodeSettings/DeveloperSettings.xcconfig`.
- Shifts the responsibility on NNW to provide the `OAuthAuthorizationClient` value since NNW just is the client and should own its own information.
- Release builds by default build for connecting to Feedly's production API. Debug builds by default build for the the sandbox API.
- Search the **workspace** for `FEEDLY_ENVIRONMENT` to find the locations to change the debug environment. It is the developers responsibility to ensure the `OAuthAuthorizationClient` ultimately matches the API environment.
- Token refresh will only occur if NNW provides the client to the account delegate sometime after the creation of the account and its delegate. I have covered the first account creation use cases but...

I need guidance (please!). When a Feedly account is restored from disk, where in the codebase should NNW give the restored `Account` the appropriate `OAuthAuthorizationClient` value?

Implements #1029.


brentsimmons:
If I understand correctly — the `OAuthAuthorizationClient` object needs to be stored with the Account, saved to disk, then reloaded on startup.

(If I don’t understand correctly, let me know!)

Here’s the scoop:

Every Account has a Settings.plist file on disk. It’s a serialized version of an `AccountMetadata` object, and it contains things like name, username, conditional GET info, and so on.

On startup, the Account sets, on its delegate, the `accountMetadata` property, so the delegate has access to the `AccountMetadata` object.

It sounds like `AccountMetadata` needs to be expanded to include an optional `OAAuthorizationClient` property — which should be Codable so it can be serialized/de-serialized to disk (to Settings.plist).

Make sense?

If I totally misunderstand the question, let me know. :)

kielgillard:
Thanks for the help and chat on Slack, @brentsimmons.

kielgillard:
To be clear, the first comment above is outdated. I've since received guidance, made changes and this PR is now ready for review and if acceptable, suitable for merging.

So since the initial comment and contrary to it, here's the final set of changes introduced in this PR:
- Bumps the minimum deployment target from 10.14 to 10.15.
- Account.framework provides an `OAuthAccountAuthorizationOperation` for NNW to create an `Account` of a certain `AccountType` that supports `OAuthAuthorizationGranting`. This operation uses `ASWebAuthenticationSession`.
- The operation makes `AccountsFeedlyWebWindowController` redundant and is therefore removed.
- Modifies the Build Phases of Account framework to substitute the production ("feedly cloud") OAuth client identifier and secret from build settings into `OAuthAuthorizationClient+Feedly.swift`. In my own testing, I added values for these build settings to `SharedXcodeSettings/DeveloperSettings.xcconfig`.
- Release builds by default build for connecting to Feedly's production API. Debug builds by default build for the the sandbox API.
- The default `OAuthAuthorizationClient` values used to connect to Feedly APIs are owned by the Account framework; the NNW module is no longer expected to inject the appropriate client value.
- Now that the Account framework owns the authorization information, `FeedlyAccountDelegate.environment.oauthAuthorizationClient` guarantees the OAuth client information matches Feedly production or sandbox API. 
- Debug builds can connect to Feedly production/cloud API instead of sandbox by changing the value returned by `FeedlyAccountDelegate.environment`. The stuff about `FEEDLY_ENVIRONMENT` above is inapplicable and removed.
- I've been given the guidance I need and token refresh is no longer broken.

PR #1286, opened 2019-11-11
Enables accounts and their delegates to prepare to be deleted. 

kielgillard:
I would like to get feedback about the approach I'm taking in this PR.

In this PR:
- Enables accounts and their delegates to prepare for deletion.
- Provides a variation of the Account Inspector on iOS for Feedly.

The use case I have in mind is this. When a Feedly account is just about to be deleted, the Feedly delegate can make a `logout` request to the API to revoke the access token issuing grant the user has previously given NetNewsWire. This way, the user is in control of what is accessing their content.

Mac PR in the works.

### DIFF
--- a/Frameworks/Account/Account.swift
+++ b/Frameworks/Account/Account.swift
@@ -399,6 +399,10 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
 		opmlFile.save()
 	}
 	
+	public func prepareForDeletion() {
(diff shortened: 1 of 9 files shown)


vincode-io:
When you can, could you please update this pull request so that it doesn't include the special Feedly account inspector?

kielgillard:
> When you can, could you please update this pull request so that it doesn't include the special Feedly account inspector?

Done.

PR #1347, opened 2019-11-25
Allow Feed Wrangler accounts on iOS

jbennett:


iOS/Account/FeedWranglerAccountViewController.swift
```diff
@@ -0,0 +1,173 @@
...
+	
+	@IBAction func showHidePassword(_ sender: Any) {
+		if passwordTextField.isSecureTextEntry {
+			passwordTextField.isSecureTextEntry = false
+			showHideButton.setTitle("Hide", for: .normal)
+		} else {
+			passwordTextField.isSecureTextEntry = true
+			showHideButton.setTitle("Show", for: .normal)
+		}
+	}
```
kielgillard: I thought you might like to know I noticed you're using `NSLocalizedString` in every other place except here.

jbennett:
Good call, thanks

Issue #1083, opened 2019-09-28
Generalize handling of secrets and keys


jbennett:
We are currently wrapping the compile build phase with a setup/reset script to replace a placeholder for all secrets/keys. This is currently running on a per file, per target, per framework basis which is tedious and seems like it should be probably be generalized into something a bit more robust.

PR #1364, opened 2019-11-27
Generalize handling of secrets and keys

jbennett:
This uses gyb + the current .xcconfig files to created a `secrets.swift` file which can be included in other targets. This file will use all of the environment variables to create properties on the `Secrets` enum so that you can do something like `Secrets.FeedWranglerClientKey`. 

This keeps secrets out of git and provides a slight amount of obfuscation if someone disassembles the binary #securityThroughObscurityFTW. It also means we have a single dynamic secrets file and don't need to have a replace/reset built script for every secret in the app.

See: https://nshipster.com/secrets
Addresses issue #1083


Shared/Secrets.swift.gyb
```diff
@@ -0,0 +1,37 @@
...
+
+  static var FeedWranglerClientKey: String {
+      let encoded: [UInt8] = [
+      % for chunk in chunks(encode(os.environ.get('FEED_WRANGLER_KEY'), salt), 8):
+          ${"".join(["0x%02x, " % byte for byte in chunk])}
+      % end
+      ]
+
+      return decode(encoded, salt: salt)
+  }
```
vincode-io: Do we have to replicate this whole block every time we need to add a new key?  Could you refactor this guy to have the encoded value generated by a reusable function?

vincode-io:
@brentsimmons This is a more secure and convenient way for us to manage secrets.

The downside is that it relies on [GYB](https://nshipster.com/swift-gyb/) which is a Python script created by the Swift core team.  We would be adding build time dependencies on Python and the GYB script to NetNewsWire.

jbennett:
I'm not surprised by that comment, lol.

If thats something we want to avoid I can tackle it a few other ways. If we don't care about obfuscating the keys (which isn't giving us much of a win) that would make it even easier.

Example of the generated secrets file:

``` Swift
// Generated by Secrets.swift.gyb
enum Secrets {

  static var feedWranglerKey: String {
     let encoded: [UInt8] = [
          0xa0, 0x34, 0x6a, 0x40, 0xee, 0xb8, 0x29, 0x5f, // …
      ]

      return decode(encoded, salt: salt)
  }

}

private extension Secrets {

  private static let salt: [UInt8] = [
    0x96, 0x07, 0x0b, 0x22, 0xd8, 0x89, 0x19, 0x6b, // …
  ]

  private static func decode(_ encoded: [UInt8], salt: [UInt8]) -> String {
    String(decoding: encoded.enumerated().map { (offset, element) in
        element ^ salt[offset % salt.count]
    }, as: UTF8.self)
  }

}
```

Shared/Secrets.swift.gyb
jbennett: This is now generated from a list of secret keys. The array `['MY_KEY', 'MY_OTHER_KEY']` generates blocks for `myKey` and `myOtherKey`

vincode-io: Nice work!  That is very cool.

vincode-io:
Could we just include GYB in the repo and reference it from the build script?  That would allow us to bundle that dependency, so that developers basically don't have to worry about it.  We still have Python installed on macOS by default for now...

jbennett:
That is a great call, thanks.

brentsimmons:
Agreed — let’s just include GYB in the repo.

PR #1367, opened 2019-11-27
Fix successive favicon URL attempts

Wevah:
The call to `self.postDidLoadFaviconNotification()` needs to be outside the brackets, otherwise `FaviconDownloader` won't try the rest of the favicon URLs on failure.


Wevah:
Wait, I need to fix a few more things that are master-specific.

PR #1368, opened 2019-11-27
Favicon master fix

Wevah:
The call to self.postDidLoadFaviconNotification() needs to be outside the brackets, otherwise FaviconDownloader won't try the rest of the favicon URLs on failure.

Also restore marking the caches dirty.


vincode-io:
Thanks for the review and the fix @Wevah.  I was coming off break and was going to ask you to review my work, but you already had it done.  Awesome.

Wevah:
There might be some more stuff missing, but I'll get to that as I notice it! (`master` seems to be working properly again with these small changes.)

Issue #1108, opened 2019-10-06
Put Debug menu behind hidden preference


brentsimmons:
The pref should be named ShowDebugMenu. It should be available in the release build, even the MAS build, and not just when running from Xcode.

The web inspector, however, should not be available in the MAS build.

correia:
I can take care of this.

jamadden:
> The web inspector, however, should not be available in the MAS build.

That seems unfortunate. I’m not familiar with the sandbox restrictions these days; I guess that’s one of the limitations?

brentsimmons:
The web inspector requires the use of private API. Apple would reject the app if we tried to include it.

Wevah:
FWIW, you should still be able to enable it (or at least the "Inspect Element" context menu item) manually by setting the appropriate user default.

Wevah:
This seems to have never made it over to `master`.

vincode-io:
Would you be good enough to submit pull request to get master caught up?

Wevah:
Sure.

PR #1372, opened 2019-11-28
Enable ShowDebugMenu on master

Wevah:
`master` version of the patch for #1108 (which AFAICT is identical; `mac-release` patch was #1110).


Wevah:
It seems like this patch might re-enable automatic reloading on debug builds, which as I understand it was specifically disabled there, but I'm not sure.

Issue #1459, opened 2019-12-24
Fix layout of code with some feeds


brentsimmons:
https://blog.emberjs.com/feed.xml

![Image from iOS](https://user-images.githubusercontent.com/1297121/71387146-47a54100-25a7-11ea-9225-911f99c88a3b.png)

Wevah:
The code blocks are rendering super-huge.

vincode-io:
This one has me stumped for now.  The problem seems to be that the `pre` tags are embedded in a `table`.  It looks like all the CSS attributes in the `pre` tag that exceeds the width of the page are ignored, including font.

The closest thing I could come up with for a fix is this:

`td > pre {
	white-space: pre-line;
}`

This forces offending `pre` block to wrap when in a `td`.  The styling no longer seems to be lost in this case, but we then have line wrapping which isn't  a desired behavior.

Wevah:
Does adding `overflow: scroll` anywhere help?

vincode-io:
We are already using `overflow: auto` on the `pre` elements.

Wevah:
Gotcha. FWIW, the font is the correct size and wraps on `mac-release`.

vincode-io:
Yeah, the CSS to do the `overflow: auto` behavior was added to the Mac 5.1 version.  Looking at `master` for the Mac, the `pre` block isn't correct, but at least the line numbers line up and the font is correct.

Wevah:
I've got a partial fix that does 2 things:

1. Adds `initial-scale=1` to the `meta name="viewport` tag (fixes initial zoom).
2. Adds `-webkit-text-size-adjust: none;` to the iOS style sheet (fixes the huge font size).

Another option would be to use JS to add a wrapper `div` with `overflow-x: auto` around any `table`s.

Wevah:
Of course, this fix means the whole feed scrolls horizontally instead of just the code table; test it out and let me know what you'd prefer.

Wevah:
Actually, I think this PR should probably go in (if it doesn't break other stuff) whether we end up wrapping `table`s or not.

vincode-io:
@Wevah That PR helped immensely.  Do you want to try the JS wrapped `DIV` with overflow to see how that works.  I think it would probably help in other situations where we have large tables as well.

vincode-io:
@Wevah Nice work!

PR #1465, opened 2019-12-25
Make initial scale 100% and don't let WebKit increase text size

Wevah:
Issue #1459.


Wevah:
The only issue with this fix is scrolling horizontally scrolls the entire feed instead of just the code table.

Issue #1449, opened 2019-12-23
iOS: Error when syncing with Feedly


marcoskirsch:
After entering my Feedly account details (iPad), NetNewsWire fetches the feeds I'm subscribed to ok – everything is populated correctly under "Feedly" in the left side pane; but never succeeds to retrieve the actual items. I get a dialog that says:

> **Error**
>
> The data couldn't be read because it is missing.
>
> ---
> Dismiss

The same exact thing happens, unsurprisingly, on iPhone.

I have version 5.0 (23).

![IMG_0103](https://user-images.githubusercontent.com/8085768/71335979-e7f95800-250a-11ea-89b2-146d69eb2d25.PNG)

gerriet:
For me it worked in iPhone, but have the same error message on iPad. Same version.

marcoskirsch:
Looks like it synchronized ok this morning on my iPad.

marcoskirsch:
Now I’m getting the error again.

kielgillard:
@marcoskirsch Thanks for creating this issue. Do you have a Mac? If so, can you please follow these instructions?

1. Connect your iOS device to your Mac.
2. On your Mac, open Console.app.
3. Using Console.app, choose the iOS device connected to your Mac in the “Devices” sectino of the side bar on the left.
4. From the menu bar, click the “Action” menu. 
5. Ensure “Include Info Messages” and “Include Debug” messages is checked.
6. In the search bar, type “Feedly”
1. On your connected iOS device, attempt to reproduce the alert.


Messages about the behaviour of the Feedly integration in NNW should be revealed in the Console as you use the Feedly account.

kielgillard:
Would appreciate it if you could try the above too please @gerriet (when you get the error!).

gerriet:
@kielgillard I tried the steps you suggested but got no log message from Feedly. NNW gave me mostly messages with "send gesture actions" and one "SecTrustReportTLSAnalytics". When I start the Feedly app I do get log messages from it.

kielgillard:
@gerriet I’ve had this same feedback about the instructions and yet the very same person got back to me today with the logs. It’s a confusing process, I know. So I have to ask: can you please try a pull to refresh or killing and launching NNW in step seven?

gerriet:
[...]

PR #1490, opened 2019-12-29
Models the feed identifier of feedly articles as an optional…

kielgillard:
…since it seems the Feedly API will not always provide one ([despite the API documentation](https://developer.feedly.com/v3/entries/#get-the-content-of-an-entry)).

This may result in NNW dropping articles since NNW requires articles to identify a subscribed-to feed.

A possibly complete fix for #1449


kielgillard:
This should be safe to merge into ios-candidate.

vincode-io:
@kielgillard Could you please submit this pull request to ios-candidate?  That way we don't have to try to cherry pick commits from master.

kielgillard:
Will do.

Issue #1501, opened 2019-12-31
Accounts unit tests fail to build


lithium3141:
In a new clone of f37ce4357e13ddc760d53608b21375f16f6945f0, I'm unable to build and run the unit tests associated with the NetNewsWire scheme. It looks like there may have been some drift in recent changes to the API under test — most of the failures tend to be due to

* missing completion blocks, or
* calling `throw`ing API without `try`, or
* the use of a Result type instead of an underlying data type.

I'd like to get these tests building again, and then run them and assess any failures separately.

lithium3141:
Brent noted I should be looking at the `ios-candidate` branch; the same issue is present in 8a85b18d097abc7bf5926061ae48e28287200ed9.

kielgillard:
I started work in this and stopped because I wanted to get the tests passing as well as building. To get them passing requires significant changes to the Feedly syncing due to other changes (such as statuses and articles being returned for subscribed-to feeds only).

Building failing tests is better than not building at all and we can always disable the failing tests until the Feedly syncing implementation catches up.

PR #1507, opened 2020-01-01
Fix Feedly tests

Wevah:
I think this should fix the Feedly tests, but I don't have Feedly credentials to check. I’m not sure if force-unwrappings that should succeed are okay here, e.g.

#1501


Wevah:
Tests will still fail, and I didn’t realize this was already actively being completed. 😩

PR #1509, opened 2020-01-01
Update Account tests for new async API

timothyekl:
This PR updates unit tests in AccountTests, primarily against the Feedly service, to accommodate new async syntax in Account.framework. A more detailed description of the issue is in #1501. After these changes, the tests build for me; I have not run them, due to not yet having a Catalina system at hand.

### DIFF
--- a/Frameworks/Account/AccountTests/Feedly/FeedlySendArticleStatusesOperationTests.swift
+++ b/Frameworks/Account/AccountTests/Feedly/FeedlySendArticleStatusesOperationTests.swift
@@ -81,8 +81,8 @@ class FeedlySendArticleStatusesOperationTests: XCTestCase {
 				let statusCount = try result.get()
 				XCTAssertEqual(statusCount, 0)
 				selectPendingCountExpectation.fulfill()
-			} catch let e {
-				XCTFail("Error unwrapping database result: \(e)")
+			} catch {
+				XCTFail("Error unwrapping database result: \(error)")
 			}
 		}
 		waitForExpectations(timeout: 2)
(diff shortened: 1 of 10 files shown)


Frameworks/Account/AccountTests/Feedly/FeedlySendArticleStatusesOperationTests.swift
```diff
@@ -74,15 +75,26 @@ class FeedlySendArticleStatusesOperationTests: XCTestCase {
...
-		XCTAssertEqual(container.database.selectPendingCount(), 0)
+		let selectPendingCountExpectation = expectation(description: "Did Select Pending Count")
+		container.database.selectPendingCount { result in
+			do {
+				let statusCount = try result.get()
+				XCTAssertEqual(statusCount, 0)
+				selectPendingCountExpectation.fulfill()
+			} catch let e {
+				XCTFail("Error unwrapping database result: \(e)")
+			}
```
kielgillard: I'm not sure if you're using the `let e` pattern deliberately or not but [this might be interesting](https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html#ID541) in case you didn't already know:
> You write a pattern after catch to indicate what errors that clause can handle. If a catch clause doesn’t have a pattern, the clause matches any error and binds the error to a local constant named error. 

Do you could write instead:
```
} catch {
 	 XCTFail("Error unwrapping database result: \(error)")
}
```

timothyekl: Hey, I didn't know about that! Thanks for pointing it out. Is this "magic" variable something that falls within this project's coding guidelines? (It reminds me of synthesized `oldValue` and `newValue` variables in property setters, but neither this `error` nor those variables are referenced in CodingGuidelines.md.)

kielgillard: I wondered the same thing. I don't think conversation has been had yet. Maybe we should get it started?

brentsimmons: It’s totally a-okay to use the magic `error` variable. I do it.

timothyekl: Filed as #1512; I'll wait for that to get resolved.

brentsimmons:
Is this PR ready to be merged?

timothyekl:
Now it is!

Issue #1485, opened 2019-12-29
iOS: Timeline context menu should have “Mark Above as Read” and “Mark Below as Read”


brentsimmons:
These should replace “Mark Older as Read.” And they should do what they say on the tin. (They should pay attention to the current sort order. “Above” could be mean older or newer, depending.)

hartlco:
I announced in the Slack room that I would be happy to work on this. Just leaving a comment so I could get the ticket assigned to me, if needed.

vincode-io:
@brentsimmons there isn't a keyboard shortcut for "Mark Above as Read".  What key should we assign that command to?

brentsimmons:
cmd-ctrl-K

PR #1514, opened 2020-01-02
Switch order of context menu items to show delete at the bottom #1503

### LINKED ISSUE #1503: Delete should be at the bottom of the context menu.
Delete should be at the bottom of the context menu, like how it is in other apps such as Safari. It’s currently in the middle. 

D601B561-90B9-4F88-A3E5-11AE9DA79775

hartlco:
I switched the order of the `rename` and `delete` action and moved the `Mark All as Read` action above them. Fixes #1503 

I was considering moving the `Mark All as Read` action to the top to be aligned with NNW for Mac but I feel this would impact usability as the action gets harder to reach for most cases.

Very happy to adjust if needed!


vincode-io:
Thanks for the pull request.  #1485 is available if you are looking for something to do next.  Just announce that you are working on it in the Slack #work channel to reserve it.

PR #1533, opened 2020-01-05
Share string with subject and URL instead of plain URL for Omnifocus and Things

### LINKED ISSUE #1517: iOS: Sharing article to OmniFocus leaves new item title blank
When I use the OmniFocus share sheet widget to share an article in NNW 5.0 (24), the URL is set properly in the notes field but the title of the article is missing from the title of the new OmniFocus item.

If I share a web page from Safari (or an article from my other newsreader, Byline) the title of the current page or article is in the title field of the OmniFocus item and the URL is in the Notes field.

hartlco:
This issue fixes #1517 

Omnifocus and Things don't call the `subjectForActivityType` method to obtain the title of a shared article. Instead they use the first line of a shared string as the title for a to-do and use the rest to populate the notes field.

For this PR I decided to specify the exceptions as to when a formatted string should be shared and sticking with sharing a plain URL as default. Many Apps provide special functionality when only a URL is shared. For example Messages.app with preview cards.

### DIFF
--- a/iOS/ArticleActivityItemSource.swift
+++ b/iOS/ArticleActivityItemSource.swift
@@ -23,7 +23,18 @@ class ArticleActivityItemSource: NSObject, UIActivityItemSource {
 	}
 	
 	func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
-		return url
+		guard let activityType = activityType,
+			let subject = subject else {
+				return url
+		}
+
+		switch activityType.rawValue {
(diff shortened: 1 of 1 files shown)


douglashill:
For context, here’s Firefox’s implementation dealing with this same problem: https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Share/TitleActivityItemProvider.swift

OmniFocus will also pick it up correctly if the URL and string (title) are provided as two separate activity items.

I’m interested if you know other apps that have special handling of only a URL.

PR #1539, opened 2020-01-06
Dragndrop feed crash fix

myurieff:


vincode-io:
I don't think that showing forbidden in the source folder is the right thing.  If someone begins a drag and changes their mind, it looks like they aren't allowed to drop the feed back where they grabbed it from.  It is better to just ignore the drop which is what the current code should be doing.

Issue #1548, opened 2020-01-07
iOS: Mark Above as Read and Mark Below as Read should not appear when there isn’t anything to mark read


brentsimmons:
It’s entirely possible that one might show and the other not show, depending on what’s currently unread in the timeline.

PR #1553, opened 2020-01-07
Only show mark unread above/below actions if articles above/below contain unread items

hartlco:
This PR dresses issue #1548 

Instead of only checking for the first/last item in canMarkAboveAsRead and canMarkBelowAsRead it's now using Array<Article>.canMarkAllAsRead to determine wether the action should be shown.

As the logic to get the articles above/below a certain article is now used canMarkAbove/BelowAsRead and markAbove/BelowAsRead I moved it to private helper functions.
I was considering moving the helper functions to `ArticleArray` but was not sure if there are special rules adding something to the shared codebase.

### DIFF
--- a/Shared/Timeline/ArticleArray.swift
+++ b/Shared/Timeline/ArticleArray.swift
@@ -109,5 +109,30 @@ extension Array where Element == Article {
 		}
 		return true
 	}
+
+	func articlesAbove(article: Article) -> [Article] {
+		guard let position = firstIndex(of: article) else {
+			return []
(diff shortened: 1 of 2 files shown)


brentsimmons:
Do we have an extension on Array where member = Article? I think we do, but I’m not in front of the code right now.

If so, then would you move the methods that get the above or below array into that extension?

hartlco:
Yes this extension exists. I will move it and update the PR after a quick test run.

PR #1556, opened 2020-01-07
Use new MacroProcessor replacement

Wevah:


brentsimmons:
Does the `!try` compile? Somehow I have it in my head that it should be try!

Wevah:
> Does the `!try` compile? Somehow I have it in my head that it should be try!

No, it doesn't. I messed up the first PR for this and in my haste I typoed and pushed in rapid succession; I caught it while reviewing the commit diffs before I saw your comment, at least!

PR #1613, opened 2020-01-12
Treat empty URLs as nil for web feed home pages

### LINKED ISSUE #1612: Empty alternate URLs should be treated as nil for home pages/favicons
E.g., in the Swift blog feed: `<link type="text/html" href="" rel="alternate"/>`. This can cause a local favicon to be used if the user is running a web server on their machine.

Wevah:
Fixes #1612.

The correct fix is probably to make `-rs_normalizedURLString` return `nil` when the URL string is all whitespace, but that would require extra code changes and testing, and can probably wait until it's ported to Swift.

### DIFF
--- a/Frameworks/Account/WebFeed.swift
+++ b/Frameworks/Account/WebFeed.swift
@@ -42,7 +42,7 @@ public final class WebFeed: Feed, Renamable, Hashable {
 			return metadata.homePageURL
 		}
 		set {
-			if let url = newValue {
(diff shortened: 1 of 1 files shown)


Wevah:
This might only visibly affect the Mac build; I’m getting the correct icon on iOS at the moment.

brentsimmons:
See also: https://github.com/Ranchero-Software/RSParser/issues/44

Wevah:
That seems like a better location for a fix.

Issue #1625, opened 2020-01-15
Give the system a chance to open universal links


kielgillard:
When tapping a universal link in an article, give the system a chance to open it in an installed app. If a universal link cannot be opened by an installed app, open it in SFSafariViewController.

kielgillard:
Born out of the discussion in https://github.com/Ranchero-Software/NetNewsWire/issues/1614. Brent's rationale can be found [here](https://netnewswire.slack.com/archives/CCGTPJDPD/p1579119670429000). The difference to #1614 is that NNW is not manufacturing universal links.

kielgillard:
Verified with Twitter and YouTube URLs.

PR #1626, opened 2020-01-15
Gives iOS a chance to open universal links

kielgillard:
This PR implements #1625. 

Turns out Twitter [doesn't seem to support universal links](https://twitter.com/.well-known/apple-app-site-association) per Apple's [documentation](https://developer.apple.com/documentation/security/password_autofill/setting_up_an_app_s_associated_domains#3001215). I'll reach out to Twitter and see what's going on.

One way I verified this PR is by installing YouTube on my device and opening a YouTube URL. For example, the kottke.org article titled "Noah Takes a Photo of Himself Every Day for 20 Years" contains a link "The Simpsons even did a parody of it" to a YouTube video.

### DIFF
--- a/iOS/Article/WebViewController.swift
+++ b/iOS/Article/WebViewController.swift
@@ -354,9 +354,17 @@ extension WebViewController: WKNavigationDelegate {
 			
 			let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
 			if components?.scheme == "http" || components?.scheme == "https" {
-				let vc = SFSafariViewController(url: url)
-				present(vc, animated: true)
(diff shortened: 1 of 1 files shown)


kielgillard:
Turns out I was wrong about Twitter: they _do_ [support](https://twitter.com/apple-site-association) universal links as per Apple's [documentation](https://developer.apple.com/documentation/security/password_autofill/setting_up_an_app_s_associated_domains#3001215) -- I just didn't see the other section in Apple's documentation where iOS can find the association file in multiple places (not just where you're supposed to add it). Once I rebooted iOS 13, it worked.

PR #1644, opened 2020-01-17
Positions the footnote popover on top of other content

### LINKED ISSUE #1611: iOS: Newsfoot footnotes display under YouTube(?) videos
B317606F-A619-432F-9FBD-E01F24602038

This Newsfoot footnote got covered up by a video...  Pretty sure it's YouTube.

pdokas:
Fixes #1611.

I always feel a little weird pulling out the "use a very high `z-index`" trick, but this markup needs to contend with anything people put into feeds so it feels reasonable to reach for a very high level.

I also thought about creating a list of CSS vars to give semantic labels to global z-index rules, but really this is the first global z-index value we’ve needed. I think it's a good idea if we ever start getting at all complex with the layering.

image

### DIFF
--- a/Mac/MainWindow/Detail/styleSheet.css
+++ b/Mac/MainWindow/Detail/styleSheet.css
@@ -212,6 +212,7 @@ img[src*="share-buttons"] {
 .newsfoot-footnote-container {
 	position: relative;
 	display: inline-block;
+	z-index: 9999;
 }
 .newsfoot-footnote-popover {
 	position: absolute;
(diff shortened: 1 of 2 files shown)


vincode-io:
@pdokas Could you submit your pull request against the ios-candidate branch please?  That is where we are stabilizing the iOS release.

pdokas:
@vincode-io Changed the base branch and rebased my branch to be off of `ios-candidate` as well.

PR #1667, opened 2020-01-20
Stop media playback when article will disappear

hartlco:
This PR addresses issue #1647 

Stopping the playback of pure video element can easily be done by calling `pause` on the element.
Youtube and Vimeo videos are wrapped in frames, there we can't directly access the video element. Resetting the `src` of the iframe will stop the playback.

I'm not sure if this is a solution we want to go for but it worked well in my testings. I also included stopping of audio element.

As usual, very happy to adjust or explore further if we can go for a different solution.

### DIFF
--- a/iOS/Article/WebViewController.swift
+++ b/iOS/Article/WebViewController.swift
@@ -123,6 +123,12 @@ class WebViewController: UIViewController {
 		}
 		
 	}
+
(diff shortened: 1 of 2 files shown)


vincode-io:
Thanks for going the extra mile to figure out how to stop the iframe video/audio.  Nice work as usual!

PR #1669, opened 2020-01-21
Update to latest RSParser

Wevah:
And remove the no-longer-needed isEmpty check in WebFeed.


brentsimmons:
Let’s leave this extra check in. I’m feeling a little paranoid.

Issue #1709, opened 2020-01-26
iOS: Wrong 'vibrant' tap-down state text colour for account type buttons


nostodnayr:
1. Settings › Add Account
2. Tap on an account type

Bonus marks: Timeline Layout button, too.

brentsimmons:
@nostodnayr I’m not seeing the wrong text color. Would you add a screenshot?

nostodnayr:
Sorry I wasn't very clear.

![Uploading 845B1418-EA0F-4229-BEBF-91058A8BAD6F.png…]()
![3575B835-8F54-4260-BB9E-6B3BEBE40E4E](https://user-images.githubusercontent.com/2190735/73143624-c11cdb80-40df-11ea-9633-2cbe06af55b1.png)

brentsimmons:
Got it! I was looking in the wrong place. Good catch.

PR #1719, opened 2020-01-27
Fix vibrant colors on highlight inside AddAccountViewController and the timeline layout cell

hartlco:
I was able to fix the issue inside `AddAccountViewController` by using cells of class `SettingsAccountTableViewCell` instead of `VibrantTableViewCell`. This enabled me to hook up to the outlets used inside the overridden `updateVibrancy` method.

I'm not sure if you prefer reusing cell classes, even though they are used for cells with different appearances, or if we should create a new cell-subclass in this case. Happy to adjust from feedback!

The issue with the timeline layout button could be solved by just using the basic style instead of the custom one. The appearance doesn't change noticeably.

This PR addresses issues #1709


vincode-io:
The cell reuse is fine. The timeline fell used to be a basic cell, but the alignment was off enough that we got a defect for it. What ever the solution we choose for the timeline cell, it should be pixel perfect.

hartlco:
@vincode-io Now diffing the image I see the alignment change. Will update the PR with a solution for the custom cell!

hartlco:
Updated the previous solution. Now it's using a custom layout for the timeline cell, just hooked up to `VibrantBasicTableViewCell`. Connecting the label outlet makes the colors change on highlight.

PR #1763, opened 2020-02-01
Don't block vertical pan gestures

### LINKED ISSUE #1760: iOS: Pull to refresh has dead zones
In the Feeds list, if the spot where your finger lands to pull down to refresh happens to be a heading (Feedly, On My iPad, Smart Feeds), the list does not pull down, so no refresh occurs. The pull has to start on an area of the list that is not gray. There shouldn't be any dead zones—pulling down anywhere in the list should work.

Wevah:
Fixes #1760.


Wevah:
Actually, this isn't quite correct. The vertical scrolling on the rest of the view has a threshold that's not simply `y > x`.

PR #1771, opened 2020-02-03
Set lastImageCacheFlushDate on launch in CacheCleaner instead of registerDefaults()

### LINKED ISSUE #1770: lastImageCacheFlushDate is never actually persisted
We set it to the current date via `UserDefaults.standard.register()` on launch, which means the cleanup branch is never hit in `CacheCleaner`, which in turn means `lastImageCacheFlushDate` is never set outside of `register()` and is never persisted to the defaults database. This means the cache is never actually purged.

Wevah:
So it's actually persisted to disk.

Fixes #1770.

(Also includes a spacing tweak in `FaviconDownloader`.)

### DIFF
--- a/Mac/AppDefaults.swift
+++ b/Mac/AppDefaults.swift
@@ -226,8 +226,7 @@ struct AppDefaults {
  		let showDebugMenu = false
  		#endif
 
-		let defaults: [String : Any] = [Key.lastImageCacheFlushDate: Date(),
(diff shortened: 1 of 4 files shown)


Wevah:
Actually, would it be better to try to wipe the cache if `lastImageCacheFlushDate` is nil? This would let a developer simply delete the defaults key to force a wipe, instead of setting the date to something in the past.

PR #1796, opened 2020-02-16
Add German localization for iOS app

J-rg:
Made some changes necessary for localizing the iOS app into other languages and localized the app into German.

Two remarks:
In Shortcuts, choosing an account still shows the unlocalized names of On My iPhone/iPad accounts.
The German equivalent of "Updated %@" has the formatter in initial position, so it should return a phrase starting with a lowercase letter for English and an uppercase letter for German. Setting the formattingContext to dynamic had no effect so I made the app capitalize the first letter of the string manually, which works but probably isn’t the most elegant solution.

I also did a full translation of the Mac app (https://github.com/J-rg/NetNewsWire/tree/master), but I’m not sure about the appropriate branch for the pull request.


seraphyn:
That was my 2nd thought, too.
After searching support for TinyTinyRSS :)

vincode-io:
@J-rg Here is an update on your pull request.  Hopefully you are following along in the #work Slack channel.  Bottom line, is that we will do localization in for Mac 5.2 and iOS 5.1 releases.  That means that once Mac 5.1 is moved to the mac-candidate branch, we will want to make our localization changes to the master branch.  

There will be guidance forthcoming that may change some of what we expect when doing localization and could impact your work.  The best way to stay on top of this is to join the #work channel in our Slack.

vincode-io:
@J-rg We are in the process of rewriting NetNewsWire in SwiftUI.  That will be the first version that we begin localization efforts on.  Thanks for all your work on this and I hope you contribute when we start localization efforts after we release NNW 6.0.

PR #1848, opened 2020-03-01
Save outline expanded state

### LINKED ISSUE #562: Folders collapse
When closing NetNewsWire 5 the smart feed folders collapse. I want them to stay open.

Wevah:
Saves expanded/collapsed state for accounts and folders, and uses the current defaults for first launch. I *think* these states will be used for new (i.e., not restored) windows with the same control identifiers once multi-window support is done, but that'll need to be verified.

The only thing that still needs implementing is restoring the state of expanded folders inside of collapsed accounts when the account is re-expanded. The built-in mechanism saves them, but doesn't auto-expand the folders.

Currently, the state will be completely collapsed for returning users, since the system's state-saving stuff tracks expanded state, not collapsed. A user default along the lines of `hasBeenRunWithOutlineStateSaving` might be appropriate.

(Mostly? fixes #562.)


Wevah:
This shouldn't be merged until we decide if we want new windows to use the last-changed expansion state, or if an explicit preference would be better. (I'm mostly in favor of the former, FWIW.)

Wevah:
Closing this PR until we decide how we want to handle the saved window stuff.

Issue #1870, opened 2020-03-10
iOS: Account Login Field should switch first responder on return


flowinho:
Currently, when the user enters his username / email into the corresponding field he has to manually select the password field.
This should happen when the user presses return.
Also the keyboard should be dismissed when the user presses the action button.

Optional: dismiss the keyboard when a view other than the text fields or the action button is touched.

Note: I already fixed this on my side. I can prepare a PR.

flowinho:
I added the functionality that the login is happening when the users presses return while the password field is first responder. This is obvious. Silly me 🤷🏻‍♂️

PR #1876, opened 2020-03-11
iOS: Account Login Field should switch first responder on return #1870

flowinho:
Contains work on issue [#1870] iOS: Account Login Field should switch first responder on return #1870.

This PR affects both the feedbin and feedly account view-controller.

This PR achieves the following:

- When the user presses the return key in the email textfield the first responder switches to the password field.
- If the user presses the return key while the password field is first responder then the login action is triggered.
- When the user presses the action button the first responder is resigned.
- Some redundant logic in the view controllers has been streamlined. Less code - less bugs.

Greetings from Germany!


vincode-io:
@flowinho We are working on 5.0.1 issues on the mac-release branch.  Could you change your pull request to point there please?

flowinho:
You've lost me now. The Pull request cannot point to Mac-release since there are heavy conflicts. I checked out the Mac-release branch but its like, totally different from the iOS-candidate branch.

vincode-io:
@flowinho I sincerely and profusely apologize for for the confusion.  ios-release is the correct branch, not mac-release.

flowinho:
@vincode-io Ah alright! I changed it back :D

Issue #1731, opened 2020-01-29
iOS: improve colors management for pure white/black favicons


sergeykrasilnik:
![9B8447C6-81DC-4C85-8897-4C53F927C75B](https://user-images.githubusercontent.com/45911681/73332127-7aa9b700-426e-11ea-88f0-e55ec5f6a226.jpeg)

vincode-io:
@sergeykrasilnik Could you do me a favor and supply that feed link for the pure white icon you ran into?

vincode-io:
We will want to assign this to Mac 5.1 when it is fixed for iOS 5.0.1.

sergeykrasilnik:
http://feeds.feedburner.com/dcurtis

flowinho:
Suggestion: 

count color distribution in a fav-icon and when _all_ values are close to white (nut pure white because very subtle greys will also be hard to see) we draw a 50% Grey (because that’s the standard for neutral colors) around it - but only in “light mode”, not in dark mode.

The same thing should be happening for almost pure black Favicons when the app is in Darkmode.

vincode-io:
We do this already for dark mode. We just need to do it for light mode too.

flowinho:
Ah! Nice! Where can I find the reference?

vincode-io:
https://github.com/Ranchero-Software/NetNewsWire/blob/master/Shared/Extensions/IconImage.swift

flowinho:
Perfect, I can work with that.

flowinho:
Feeds to test pure black icons (to verify that old functionality is still working):
https://500ish.com/feed

sergeykrasilnik:
How about severe cases like this http://cherenkevich.com/feed/ pure black square icon that fills all the space?)

flowinho:
@sergeykrasilnik That's really an almost impossible edgewise. Either we zoom the icon and display it, or we do nothing at all. The issue here at hand is how colours work: If we find out that the overall image is pure black, this does not mean we know it's _content_. A PNG with black pixels and some transparent background would still evaluate as _pure black_ but should nevertheless gain some background color. We _could_ examine the content of the feed-icon but that's a lot of engineering for a massive edge-cases of someone choosing a black square that fills all pixels as his feed icon.

sergeykrasilnik:
I understand that and also that this is almost a single case problem but still an interesting design task)

[...]

PR #1892, opened 2020-03-13
Add background to pure white icons

flowinho:
This PR contains work related to issue #1731 

- I extended the shared logic with "isBright"
- I added the switch to the iOS UI logic

What I *did not* do is add the logic to macOS. I think @Wevah is in charge here?


flowinho:
This PR features an issue I wasn't aware of. I'll notify you once I fixed it.

flowinho:
Done.

PR #1951, opened 2020-03-27
upgrade URLs to HTTPS

vszakats:


vincode-io:
@vszakats Could you let me know what were you trying to accomplish with this pull request?  Most of the changes were to commented code and documentation.  I'm just trying to figure out if I am missing something.

Wevah:
I think OP means if a user pastes an HTTP URL into the new subscription sheet, it should try HTTPS first.

vszakats:
Main point was to fixup default feeds to use HTTPS. These cannot be edited via the UI, only deleted and recreated manuall. And, instead of doing it manually I run the complete source tree (except tests) through my HTTPS checker/upgrader tool which fixed the rest of them as well. I can only see 4 lines that are comments (in `ArticlesTable.swift`), the rest is feeds, actual clickable links on the UI and/or documentation.

vszakats:
Pushed an update to reset `secure.flyingmeat.com` to `flyingmeat.com`. This was auto-upgraded based on latest `https-rulesets.org` rule, which is apparently a little bit out-of-date.

vincode-io:
@vszakats The only place that the default feeds are defined is in `DefaultFeeds.opml`.  Let's get a pull request with only that file and see where we are.

vszakats:
I don't understand why does it hurt to fix links also in the about box. I'm passing.

vincode-io:
What made me nervous about the pull request is that it was done globally and it didn’t look like the changes were well understood or tested.

vszakats:
If that's a thing, I've been upgrading URLs in Homebrew (for example) since 2015, using this tool for almost the whole period. Not saying mistakes cannot happen, but this patch looked fairly low-risk. Also easily verifyable to clear any doubts.

PR #1977, opened 2020-04-10
First-run at some changes to attempt to fix Issue 1976

r-token:
I only made a few changes, but I don't feel super great about them - especially the line I commented out in MasterFeedViewController (line 477).

I don't expect this to be the correct fix, but this combination of changes has stopped the crash (but an odd bug still happens occasionally if you time things perfectly).

I'm mostly just very interested in your feedback. This was my first dive into the code at all, and I really enjoyed it. Looking forward to knocking this out.

### DIFF
--- a/iOS/SceneCoordinator.swift
+++ b/iOS/SceneCoordinator.swift
@@ -723,7 +723,30 @@ class SceneCoordinator: NSObject, UndoableCommandRunner, UnreadCountProvider {
 		}
 		
 		currentFeedIndexPath = indexPath
-		masterFeedViewController.updateFeedSelection(animations: animations)
+		
+		// I know this is a bad pyramid of doom - would like to know how to best improve this
+		if let ip = currentFeedIndexPath, let node = nodeFor(ip) {
+			if let parentNode = node.parent {
+				if let containerIdentifiable = parentNode.representedObject as? ContainerIdentifiable {
+					if let selectedItem = containerIdentifiable.containerID {
+						if !expandedTable.contains(selectedItem) {
+							completion?()
+							return
+						}
+					}
+				}
+			}
+		} else {
+			completion?()
+			return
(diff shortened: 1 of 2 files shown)


brentsimmons:
I’ll let you and @vincode-io work on this — but I want to thank you for diving into this bug. Thanks!

iOS/SceneCoordinator.swift
```diff
@@ -723,7 +723,30 @@ class SceneCoordinator: NSObject, UndoableCommandRunner, UnreadCountProvider {
 		}
 		
 		currentFeedIndexPath = indexPath
-		masterFeedViewController.updateFeedSelection(animations: animations)
+		
+		// I know this is a bad pyramid of doom - would like to know how to best improve this
+		if let ip = currentFeedIndexPath, let node = nodeFor(ip) {
```
vincode-io, line 728: You could change this into a single `if` or `guard` and format the conditions so that they are stacked to prevent all the indenting.

vincode-io:
My hypothesis was wrong.  We are actually getting the collapse and select in a different order than I thought it was, which is causing your nested `if` statement to never catch the errant scenario.  It looks like you went through some heroic efforts to make this work.  Sorry about that.

My next thought is that we need to make `UITableView.selectRowAndScrollIfNotVisible` safer.  Let's make sure that the `indexPath` exists before trying to select or scroll in `selectRowAndScrollIfNotVisible`.

https://stackoverflow.com/questions/29516638/how-can-i-check-if-an-indexpath-is-valid-thus-avoiding-an-attempt-to-scroll-to

Let's close this pull request and see if this other approach works any better.

vincode-io:
You might run into some undesirable side effects like you did on the last one.  Try to catch me in Slack to discuss them before you spend more than a few minutes on them.

PR #1979, opened 2020-04-10
Removed an exclamation point? I think this completely fixes the bug?

r-token:
Take 2: I originally created this as an actual pull request instead of a draft.

Some notes:
- Reverted the changes from my first PR
- Tried a few different ideas for fixing the crash within UITableView-Extensions.swift - mostly related to checking if the IndexPath exists before selecting/scrolling
- The index path always existed, so these checks weren't working

**Fix**: removed an exclamation point from line 20. I haven't found any side effects (yet?)

Comments:
Just by reading the code, I don't fully understand how this worked with the exclamation point in there. It makes it seem like if the visibleIndexPaths don't contain indexPath, or if the cell isn't completely visible, it's fine to scroll.

By removing that exclamation point, we're ensuring that both cases are met before scrolling. The scrollToRow is what causes the crash when a row is selected while the section is collapsing. Looking forward to your feedback.

### DIFF
--- a/iOS/MasterFeed/MasterFeedViewController.swift
+++ b/iOS/MasterFeed/MasterFeedViewController.swift
@@ -75,6 +75,12 @@ class MasterFeedViewController: UITableViewController, UndoableCommandRunner {
 		super.viewWillAppear(animated)
 	}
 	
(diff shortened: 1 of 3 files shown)


vincode-io:
Closing this one as it breaks our scroll into view functionality.

PR #2013, opened 2020-04-24
Add title for AddAccountViewController

### LINKED ISSUE #2011: "Add Account" screen in settings is untitled
iOS
Version 5.0.1 (41)

When you go to Settings: Add Account, the nav bar should be labeled "Add Account". Instead, it has no label at all.

Reference: iOS Settings app: Passwords & Accounts → Add Account

rizwankce:
Fixes #2011

### DIFF
--- a/iOS/Settings/Settings.storyboard
+++ b/iOS/Settings/Settings.storyboard
@@ -519,10 +519,10 @@
             </objects>
             <point key="canvasLocation" x="284" y="151"/>
         </scene>
-        <!--Add Account View Controller-->
+        <!--Add Account-->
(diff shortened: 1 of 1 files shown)


rizwankce:
I skipped updating the title via storyboard because it might be pain when we localise the app. But feel free to let me know if storyboard approach is what needed.

vincode-io:
We are going to have to localize the storyboard stuff anyway.  Let's go ahead an make the change storyboard since that is usually where we set the title anyway.

rizwankce:
@vincode-io made the change. Title is being set by storyboard now 😅

PR #2016, opened 2020-04-24
WIP: Title tags

### LINKED ISSUE #1964: Mac: HTML rendered from an article title
image

It's a bit surprising.

I'm not sure if this is expected behaviour.

Wevah:
Leaves certain tags unencoded so they display in article titles. Also adds attributed titles for timeline cells so those tags display there as well.

The sizing based on the attributed title may be unnecessary in most cases, since most titles have no styling tags.

Will fix #1964.


Mac/MainWindow/Timeline/Cell/NSAttributedString+NetNewsWire.swift
```diff
@@ -0,0 +1,134 @@
...
+typealias FontDescriptor = UIFontDescriptor
+typealias Color = UIColor
+
+private let boldTrait = UIFontDescriptor.SymbolicTraits.traitBold
+private let italicTrait = UIFontDescriptor.SymbolicTraits.traitItalic
+private let monoSpaceTrait = UIFontDescriptor.SymbolicTraits.traitMonoSpace
+#endif
+
+extension NSAttributedString {
+
```
vincode-io, line 30: It looks like this file is common code, but is in the Mac folder.  Could we move this to Shared/Timeline?

Wevah: Oh, yes, good catch.

Wevah: I moved it into Shared/Extensions instead of Shared/Timeline.

Issue #2049, opened 2020-05-06
Folder management of accounts needs to be optional


jbennett:
Some sync services don't support folders so management of folders in NNW should be optional. I propose adding an `AccountBehavior` (`.disallowFolderManagement`) to remove the account from the relevant UI areas.

- [x] add AccountBehavior
- [x] remove from list of options in create dialog (macOS)
- [x] remove from list of options in create dialog (iOS)
- [x] remove from list of options in create dialog (iPadOS)

PR #2050, opened 2020-05-07
Adds an account behavior to disable folders

jbennett:
Some sync services (Feed Wrangler) don't support folders. This behavior removes the account from the UI when adding new folders.

#2049


jbennett:
Looks like this issue can be closed

PR #2063, opened 2020-05-11
Adds "Find in Article" activity to the share sheet

### LINKED ISSUE #1750: Being able to search through post
It would be nice to be able to open a post, click on the box with the arrow, and search for specific words in a post, just like in Safari.

bdougsand:
closes #1750

This PR adds a “Find in Article” activity to the article share sheet on iOS. Searching and highlighting is performed with JavaScript.

Specific concerns:
- `ArticleSearchBar.swift` builds that view programmatically. I'm sure I'm missing a better approach...
- While a search is active and the keyboard is hidden, the search bar takes the place of the toolbar (modeled after Safari). Unfortunately, there's a gap between the base of the search bar and the edge of the screen. Is there a correct way to temporarily replace the toolbar?
- There are some unintended changes to `Main.storyboard`, mostly in the resources section. I tried to undo them manually, but Xcode changed all the locks to my apartment and wouldn't let me back in until I reverted the fixes.

Misc:
- The JS returns all the results back to Swift, including match groups and bounding rects in client-relative coordinates, so highlighting could be done with CALayers instead of injected div elements. I believe CALayers would offer more compositing modes.
- The bounding rects could I think also be used to implement smooth scrolling on the Swift side. WebKit's `scrollBy` method doesn't take a `behavior` option, so for now I've just had the window jump to the desired position. Using a pure-JS animation would be easy to do badly.


bdougsand:
I tracked down a couple crasher bugs and prevailed upon them to change their ways. Specifically, the stack overflow noted by @vincode-io and a hierarchy inconsistency error that I was able to trigger by rotating my phone (though not the simulator). Both were because I don't really know what I'm doing and thought I could get away with abusing `inputAccessoryView`. (Spoiler: I couldn't.)

There are additional, mostly minor improvements, for which see the commit messages.

PR #2073, opened 2020-05-14
Keyboard shortcuts for "Open in App Browser"

### LINKED ISSUE #2062: iOS: Need keyboard shortcut for open in SFViewController

rizwankce:
Fixes #2062 

Works on both Feeds and Timeline list.

### DIFF
--- a/iOS/SceneCoordinator.swift
+++ b/iOS/SceneCoordinator.swift
@@ -12,6 +12,7 @@ import Account
 import Articles
 import RSCore
 import RSTree
+import SafariServices
 
 enum PanelMode {
 	case unset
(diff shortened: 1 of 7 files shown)


iOS/SceneCoordinator.swift
```diff
@@ -1183,6 +1184,21 @@ class SceneCoordinator: NSObject, UndoableCommandRunner, UnreadCountProvider {
...
+			rootSplitViewController.viewControllers.last?.present(vc, animated: true)
+		}
+	}
+
+	func showSafariForCurrentArticle() {
+		guard let preferredLink = currentArticle?.preferredLink, let url = URL(string: preferredLink) else {
+			return
+		}
+		let vc = SFSafariViewController(url: url)
+		rootSplitViewController.viewControllers.last?.present(vc, animated: true)
```
rizwankce, line 1200: Not sure this is the right way of doing it.

vincode-io: I think I would add a function in ArticleViewController (that would then call WebViewController) that can be called to open the SFViewController.

vincode-io: Instead of `openInSafari` let's use `openInAppBrowser` instead.  I think that makes it clearer that `openInBrowser` opens externally and `openInAppBrowser` opens the internal browser.

rizwankce: Sound good! let me change it

rizwankce:
@vincode-io made the requested changes

PR #2074, opened 2020-05-14
"Mark All Read" shortcut for feeds list

### LINKED ISSUE #2056: Shortcut for Mark All as Read doesn’t work when focus is on Feeds list
# Summary:
The keyboard shortcut for _Mark All as Read_ doesn’t work when focus is on Feeds list. It works when focus is in the Articles list. In both views, other keyboard shortcuts (such as _Go to Previous Feed_, _Go to Next Feed_, and _Mark All as Read and Go To Next Unread_) work.

## Steps to Reproduce:
1. Launch NNW in landscape on iPad with a keyboard attached.
2. Select a feed from the list on the left.
3. Press `K` to Mark All as Read.

## Results:
This doesn’t work. I typically use `A`/`Z` to move up and down between the list, and mostly use `L` to Mark All as Read and Go To Next Unread. However, there are times where I just want to mark all in the current feed as read but `K` does not work in this view.

## Workaround:
Press space bar to switch focus to Articles (from Feeds). `K` works here, as does `L`/`A`/`Z`.

## Regression:
I can’t tell exactly when this regressed but it does not work in _NNW 5.0.1 (46)_ and I’m fairly confident that it did work at some point.

rizwankce:
Fixes #2056 

For Feeds list short cut is "k" and for time line its is "cmd+k"

### DIFF
--- a/iOS/MasterFeed/MasterFeedViewController.swift
+++ b/iOS/MasterFeed/MasterFeedViewController.swift
@@ -458,7 +458,18 @@ class MasterFeedViewController: UITableViewController, UndoableCommandRunner {
 			self.reloadAllVisibleCells()
 		}
 	}
-	
+
+	@objc func markAllAsRead(_ sender: Any) {
+		guard let indexPath = tableView.indexPathForSelectedRow, let contentView = tableView.cellForRow(at: indexPath)?.contentView else {
+			return
+		}
+
+		let title = NSLocalizedString("Mark All as Read", comment: "Mark All as Read")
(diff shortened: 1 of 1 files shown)


vincode-io:
What does this mean?

> For Feeds list short cut is "k" and for time line its is "cmd+k"

Both "k" and "cmd+k" should work in both views.  Is this not the case?

rizwankce:
@vincode-io ya you are correct. Didn't notice that. Just tested on iPad and its working as expected

Issue #1778, opened 2020-02-06
Cmd-click should open in Safari without switching


MtnBiker:
I don't know if this is a bug or feature, but Cmd-clicking on an article title should open the Safari window behind NNW. This is a common action in macOS.

Feature: An option for right arrow from an article to do the same as above, i.e., open the article in Safari behind NNW. Has to be an option or accidental arrowing would be frustrating to those who hadn't selected the option.

Safari should of course be taken to mean default browser.

Keep up the good work.

Wevah:
There's an option in the preferences for opening in the background; Command-click should probably invert this setting.

spodick:
More of an FYI - may be Apple's or Google's problem instead.  I use the "Open in background in browser" preference.  That works fine for Safari and Firefox, but if Chrome is set as my default web browser then command-clicking a link does not open in the background and takes me directly to the link in Chrome.
MacOS: 10.15.5
Chrome: Version 83.0.4103.116 (Official Build) (64-bit)

brentsimmons:
Chrome bug.

brentsimmons:
It’s a bug in Chrome, and I don’t know of any way to work around it.

nskh:
Sounds good, thank you!

timharek:
This is also the case with Firefox 87.

When I'm using Newsboat in the terminal, I'm able to open articles from my feeds without bringing the application to the foreground with `open -ga`.

I'm not very familiar with macOS development, so I'm not sure if there is an equivilant in Swift when opening an application from another.

vincode-io:
This has been implemented, so I'm closing this one out.

Issue #561, opened 2019-02-28
There should be an Open in Browser in Background menu variant


peternlewis:
There should be an Open in Browser in Background menu variant, perhaps ⌘⇧Right Arrow?

peternlewis:
The shift-key behaviour should be toggled by the Open in Background preference.

AJK18:
I prefer the old method - it was a setting, and then it would let me just open in the background, as I prefer

JayBrown:
In Vienna it's a setting: "Open links in external browser" plus "Open new links in the background", which is very helpful, and it would be my preferred solution for NNW.

brentsimmons:
Note: this is done on master.

Issue #559, opened 2019-02-28
Command-key to toggle Open in Background preference


peternlewis:
It would be nice if clicking a web link with the command-key toggled the Open in Background setting (so clicking the link could open in normally, and command-clicking could open it in the background).

brentsimmons:
Good point. I wonder if the callback API tells me anything about modifier keys. (I suppose I could inspect the current NSEvent.)

peternlewis:
I suspect it would be safe enough to just use NSEvent.  It's unlikely to be that time critical.

Wevah:
This should be fixed in #2100 (for Mac 5.1); feel free to reopen if it's not.

PR #2100, opened 2020-05-21
Make shift key invert the "open in background" preference

Wevah:
This checks for the shift key in `NSApp.currentEvent` for many methods, and also makes it work in the web view.


Wevah:
Actually, this method might complicate some stuff being done for the keyboard shortcut elsewhere.

Wevah:
Fixes #1778, #561, #559,

Wevah:
Command-click should probably work also, but that will require special handling for when the comamnds are triggered from menu items.

PR #2112, opened 2020-05-27
Fix issue where pressing [return] would not open article in browser

MartinJNash:
- Keyboard shortcut [return] was calling selector `openFeedInAppBrowser:` which is not implemented in the macOS app. Changed to `openInBrowser:`.
- Updated keyboard shortcut title to match naming of other shortcuts that performed the same action.


rizwankce:
Guess it’s broken via my commit to iOS app. We added support for iOS app to open in app safari view controller. I’ll fix it for iOS app

PR #2121, opened 2020-05-29
fix open in app browser for mac and iOS

rizwankce:


Mac/MainWindow/MainWindowController.swift
```diff
@@ -273,6 +273,11 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 		openArticleInBrowser(sender)
 	}
 
+	@objc func openFeedInAppBrowser(_ sender: Any?) {
```
vincode-io, line 276: Could we change this to `openInAppBrowser`?  It isn't really opening the feed, it is opening the article.

rizwankce: Sure it was conflicting between iOS [return] key. I will make the changes everywhere

rizwankce: I made changes for iOS and found for Mac feed list was not opening in browser fixed that as well.

So for Mac and iOS - both feed list and articles list - both `b` and `[return]` key works as expected.

iOS/KeyboardManager.swift
```diff
@@ -178,7 +178,7 @@ private extension KeyboardManager {
 		keys.append(KeyboardManager.createKeyCommand(title: openInBrowserTitle, action: "openInBrowser:", input: UIKeyCommand.inputRightArrow, modifiers: [.command]))
 
 		let openInAppBrowserTitle = NSLocalizedString("Open In App Browser", comment: "Open In App Browser")
-		keys.append(KeyboardManager.createKeyCommand(title: openInAppBrowserTitle, action: "openInAppBrowser:", input: "\r", modifiers: []))
+		keys.append(KeyboardManager.createKeyCommand(title: openInAppBrowserTitle, action: "openArticleInAppBrowser:", input: "\r", modifiers: []))
```
vincode-io, line 181: I think there is still some bad communication on my part.  Sorry @rizwankce.

The return key should not be mapped at all in the `KeyboardManager`.  That will prevent future remapping using the `GlobalKeyboardShortcuts.plist`.  You should be able to do everything you need to do without touching `KeyboardManager`.  If there is some reason you feel like you need to do it here, please let me know.

vincode-io: `GlobalKeyboardShortcuts.plist` is intended for keymapping that in the future can be remapped using some kind of customization in the settings.  The one's that won't be customizable are the ones on macOS that are established using menu items or on iOS in the `KeyboardManager`.

Since the return key isn't mapped using a menu item on macOS and only mapped using `GlobalKeyboardShortcuts.plist`, it should be the same on iOS.  Only mapped using `GlobalKeyboardShortcuts.plist` and not mapped using `KeyboardManager`.

The difference between opening a Feed and opening a Article should depend on what the current first responder is, So they should be sharing the same selector name.

rizwankce: @vincode-io I got your point now. I refactored iOS app to rely only on Global Shortcuts file instead of keyboard manager.

vincode-io: Thank you and thank you for hanging in there with me.  It is sometimes hard to communicate this stuff over the internet.  🙂

rizwankce: Totally I get it :) I was wondering why we have plist file in first place but now I understand its because of we share it with Mac app.

PR #2134, opened 2020-06-07
Disable selection while applying snapshot

### LINKED ISSUE #2132: Marking a feed as read, then quickly tapping a different feed navigates to the wrong feed
Seems like an issue with updating the index paths of the table before it gets updated in the UI, then quickly selecting a new item in the list navigates to the one underneath.

Here's an example screenshot of my feeds:

If I were to go into `Dev` for example, then mark all as read, when the navigation stack is popped, `Dev` remains visible, but quickly disappears because I have it filtered to unread items. If I select a different row before it disappears, it navigates to the wrong row.

I will try to catch it through a screen recording to attach to this issue

Mackarous:
This fixes #2132 

Not sure the general feeling about this one, it's a quick fix that might have a more elegant solution...

Also not sure what the preferred way of adding functionality is, I made a couple private funcs around this


vincode-io:
@Mackarous I've desk checked this and it looks ok to me.  Since this will be for the 5.0.2 release, could you make your pull request against the ios-release branch?  That will save me from having to cherry pick it off of master.

Mackarous:
@vincode-io done

PR #2159, opened 2020-07-01
Mark all as read context menu

### LINKED ISSUE #2158: On Feeds screen, Mark All as Read is not available in the contextual menu for accounts

rizwankce:
Fixes #2158 

The reason why it's failing to show context menu for "Mark All as Read" in accounts is because its expecting a content view and since it doesn't have one from `UIInteraction` its failing.

So I did some refactoring and added content view from both index paths or straight away from any `UIView`.

I did test all the contextual menus on `MasterFeedController` just to make sure I don't break anything.


vincode-io:
Thanks @rizwankce.  Nice work!

PR #2161, opened 2020-07-01
AppSettings

stuartbreckenridge:
To the tune of Park Life:

All the Settings,
So many Settings,
They all go hand in hand...


vincode-io:
We should go ahead, bite the bullet, and turn `AppDefaults` into an instance object singleton, like you did for `AppSettings`.  On Multiplatform that could be the `ObservableObject`.  That way we shouldn't need `AppSettings` at all.

I realize that means refactoring some code in macOS, iOS, and Shared.  I think it is better that we just get it out of the way now rather than maintaining two almost identical classes in Multiplatform.

vincode-io:
That is looking a lot better and what I think we needed to do.  The only problem is that now we need to update the macOS and iOS apps so that they will compile again.  Their AppDefaults classes also need to be turned into singletons, but they won’t need to implement ObservableObject.

The plan is to merge the swiftui branch into master, hopefully very soon.  That way we won't get too far out of sync with ongoing bugfixes.  We'll need to be able to maintain all 3 UI frameworks along side each other to do that.

stuartbreckenridge:
Mac, iOS, and Mutliplatform are now compiling.

PR #2178, opened 2020-07-02
SwiftUI Appearance Settings

### LINKED ISSUE #2172: Implement iOS Settings Appearance Section

rizwankce:
closes #2172

### DIFF
--- a/Multiplatform/iOS/AppDelegate.swift
+++ b/Multiplatform/iOS/AppDelegate.swift
@@ -407,13 +407,13 @@ private extension AppDelegate {
 	}
 
 	func updateUserInterfaceStyle() {
-		switch AppDefaults.shared.userInterfaceColorPalette {
-		case .automatic:
-			window?.overrideUserInterfaceStyle = .unspecified
-		case .light:
-			window?.overrideUserInterfaceStyle = .light
-		case .dark:
-			window?.overrideUserInterfaceStyle = .dark
-		}
+//		switch AppDefaults.shared.userInterfaceColorPalette {
+//		case .automatic:
+//			window?.overrideUserInterfaceStyle = .unspecified
+//		case .light:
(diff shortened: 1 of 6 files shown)


Multiplatform/iOS/AppDelegate.swift
```diff
@@ -388,3 +391,29 @@ private extension AppDelegate {
 	}
 	
 }
+
+private extension AppDelegate {
+	@objc func userDefaultsDidChange() {
+		updateUserInterfaceStyle()
+	}
+
+	var window: UIWindow? {
```
rizwankce, line 400: couldn't find a better way to get reference to window. Since delegate don't have one. If you find better option, please do let me know

vincode-io: In SwiftUI I think we update the Environment to handle this.  I think changing the environment in MainApp is the best place to try at first.  https://www.hackingwithswift.com/quick-start/swiftui/how-to-detect-dark-mode

rizwankce: I did stumble upon that `Environment` property wrapper but I couldn't find anyway to set the color scheme.
Also, with provided environment color scheme, we don't have options for "Automatic" - means use the system wide settings.

Guess Environment property wrapper is only used to get the color scheme and act upon that and not to change it. But I may be wrong here.

This is what the documentation says about it

>Setting the `colorScheme` environment value directly is an advanced use
case, as it changes the color scheme of the contained views but *not* of
the container. Instead, consider using the
``View/preferredColorScheme(_:)`` modifier, which propagates to the
presentation containing the view.

vincode-io: Will there be a problem implementing this using the `preferredColorScheme` modifier?

PR #2210, opened 2020-07-04
Implement iOS progress view

philviso:
### Description

Implemented iOS progress view

### Notes

* I made `RefreshProgressModel` owned by `MainApp`, because every window in the app should share the same refresh progress state. I could see the need for a `MainAppModel` in the future, but I didn't go so far as to add that.
* I don't have a machine running the Big Sur beta, so I have no idea if this caused any issues.
* @vincode-io I added `RefreshProgressModel` environment object to the Mac scene, and specific sizing for the Mac implementation of the progress view. Do you want me to keep that in or would you prefer if I remove that until the Mac part is implemented? Looks like it's up in the air if we'll be able to use the SwiftUI toolbar on the Mac.


Multiplatform/Shared/AppAssets.swift
```diff
@@ -69,7 +69,7 @@ struct AppAssets {
 	}()
 	
 	static var faviconTemplateImage: RSImage = {
-		return RSImage(named: "FaviconTemplateImage")!
+		return RSImage(named: "faviconTemplateImage")!
```
philviso, line 72: Fixes a crash

Multiplatform/Shared/Sidebar/RefreshProgressModel.swift
```diff
@@ -0,0 +1,101 @@
...
+import Combine
+import Foundation
+import SwiftUI
+
+class RefreshProgressModel: ObservableObject {
+	
+	enum State {
+		case refreshProgress(Float)
+		case lastRefreshDateText(String)
+		case none
```
philviso, line 19: I specifically added a `none` case here instead of having the `state` variable be optional, because I didn't want the view to need to interpret what a nil state meant. I'm not sure it's even technically possible. When the app first launches a feed refresh is triggered, and whether it succeeds or fails we always end up we a last refresh date right after that. We should always be in a state of having `refreshProgress` or `lastRefreshDateText`.

Multiplatform/Shared/Sidebar/RefreshProgressModel.swift
```diff
@@ -0,0 +1,101 @@
...
+		return formatter
+	}()
+		
+	private let lastRefreshDate: () -> Date?
+	private let combinedRefreshProgress: () -> CombinedRefreshProgress
+	
+	private static let lastRefreshDateTextUpdateInterval = 60
+	private static let lastRefreshDateTextRelativeDateFormattingThreshold = 60.0
+	
+	init(lastRefreshDateProvider: @escaping () -> Date?,
```
philviso, line 37: I tried to keep this model as lightweight as possible instead of directly requiring `AccountManager` to supply its data. I planned to write tests, but it looks like the project doesn't have test targets for the multi-platform stuff yet.

Issue #2164, opened 2020-07-01
Implement multiple selection in Sidebar


vincode-io:
I can't find any native SwiftUI support for this right now.  I'm going to move forward with single selection for now, but we should implement this as it is a macOS feature currently.  Here is an article that may help implement this.  https://www.pawelmadej.com/post/multi-select-picker-for-swiftui/

stuartbreckenridge:
I have multiple selection working to a point.

I added `@Published var selectedSidebarItem = Set<Int>()` to `SidebarModel`, then updated the `List` initialiser in `SidebarView` to include it `List(selection: $sidebarModel.selectedSidebarItem)`, then added tags to each of the sidebar items in the sidebar: either `.tag(sidebarItem.id.hashValue)` or `.tag(sidebarItem.feed!.feedID!.hashValue)`. With this, you can multi-select anything, top-level items, feeds, folders, feeds in folders.

The problem is that only the Smart Feeds and On My Mac top-level sidebar items are updating the `selectedSidebarItem` Set with new hash values. You can see both hash values for top-level items in the Set when both are selected, but it updates with nothing new when you select an actual feed.

PR #2232, opened 2020-07-08
Multi-selection on macOS #2164

stuartbreckenridge:
This commit (for #2164):

• Enables multi-selection on the macOS sidebar and suppresses it on iPad OS.
• The first selection doesn't work, but subsequent selections can be tracked via `SidebarSelectionModel`.
• Print statements have been left in on `SidebarSelectionModel` to monitor behaviour.
• The commit is non-destructive...navigation as it currently works is not disturbed.


vincode-io:
Is there a reason that we need `SidebarSelectionModel`?  Shouldn't it just be a part of `SidebarModel`?  One thing we will need to do later is programmatically change the selection of the sidebar for things like automatically selecting a feed when it is added or the next unread functionality.  That will be driven by `SceneModel` and delegated to `SidebarModel`.

stuartbreckenridge:
I have moved selection to `SidebarModel`.

PR #2241, opened 2020-07-09
OPT: Cleanup Add Folder view on Mac, change SceneNavigationView to su…

alexcfaber:
…pport multiple sheet types.


alexcfaber:
Adding Folder and Adding Webfeed works from the Mac App too now.

PR #2274, opened 2020-07-17
Adds timeline preview to viewing pane / improves dark/light changes on iOS

stuartbreckenridge:


Multiplatform/Shared/AppDefaults.swift
```diff
@@ -134,7 +134,9 @@ final class AppDefaults: ObservableObject {
 		}
 		set {
 			AppDefaults.store.set(newValue.rawValue, forKey: Key.userInterfaceColorPalette)
-			objectWillChange.send()
+			DispatchQueue.main.asyncAfter(deadline: .now() + 0.5, execute: {
```
stuartbreckenridge, line 137: This delay makes the transition between modes a little more natural, especially on iOS, and it happens just after the navigation transition.

Issue #2133, opened 2020-06-07
AppleScript: add feed property to article


brentsimmons:
This way you can get the feed from the current article, for one thing.

Wevah:
Looks like `webFeed` doesn't return its fully-qualified reference (i.e., no parent chain).

Wevah:
Ok, nope `self.container` is coming back as `Application` for `selected article`'s feed.

Wevah:
This is complicated by the fact that a feed can be in multiple folders/top level account + folder.

Wevah:
Actually, maybe the full chain isn't necessary as long as the Account can find a feed by id even in folders.

Wevah:
When a feed has the same ID but is in two different places, is it backed by the same exact data?

brentsimmons:
Yes, it is.

Wevah:
It might be ok for `webFeed id "someid" of account "OnMyMac"` to find a feed in a subfolder, then.

vincode-io:
@Wevah it looks like you were working on this.  Should it be assigned to you?

Wevah:
Yes, that's fine. Still need to sort out specifics.

Wevah:
Fixed in #2294. Open new issues if extra functionality is required!

PR #2294, opened 2020-07-23
AppleScript: Add a "feed" property to articles

Wevah:
WIP for #2133.


Wevah:
Feeds can appear in multiple folders, so getting a definitive parent is tough.

PR #2300, opened 2020-07-26
Fix GitHub URLs that point to `master` instead of `main`

noelleleigh:
This repo uses `main` as its default branch name instead of `master`, but there are many URLs in the codebase that reference `master` and thus lead to 404 pages.

This PR fixes those URLs to point to `main` instead.


vincode-io:
@noelleleigh Thanks for catching this!   I'll be cherry picking this commit and putting it in mac-release and ios-release.

Issues tracking this:  #2301 & #2302

PR #2314, opened 2020-08-02
Fixes #2058: Defer request Notifications permissions on macOS until necessary

### LINKED ISSUE #2058: macOS: Don't ask for Notifications permissions until necessary
We shouldn't ask for notifications permissions until the user toggles a feed for "Notify About New Articles".  This will also give us a chance to see if they have notifications disabled and provide an error message if so.

AngeloStavrow:
Fixes #2058.

This PR essentially replicates the iOS app's behaviour in the macOS app. However, it departs slightly from the approach taken by the iOS code to prevent a race condition between _getting_ and _checking_ `userNotificationSettings` when the notification checkbox is toggled in the web feed inspector.

The iOS app calls `updateNotificationSettings()` in the web feed inspector's `viewDidAppear()` method, which fires when we dismiss a modal to return to the view (or round-trip over to the Settings app and back to NetNewsWire). On the Mac, it doesn't.

Simply calling `updateNotificationSettings()` at the top of the `isNotifyAboutNewArticlesChanged(_:)` action isn't sufficient, since fetching the user's notification settings happens asynchronously, so by the time we've retrieved the `authorizationStatus`, the code has already moved on to checking what that (now stale) `authorizationStatus` is.

To get around that, then, we first call `UNUserNotificationCenter.current().getNotifications()` to get the settings, and then run the checking logic and flip the relevant bits in that method's the completion handler.


vincode-io:
You might have missed the discussion in the Slack work channel, but we are now doing changes to Mac 5.1 on the `mac-candidate` branch.  Could you submit your pull request against the `mac-candidate` and I'll merge your pull request and then merge those changes to `main`.

AngeloStavrow:
Oops! Yeah, I missed that, sorry. Let me see if I can just edit this PR for a clean merge into `mac-candidate` — if not, I'll close and create a new one.

AngeloStavrow:
Oof, 342 files changed? Right, I'm going to close this PR and open another one, just to keep things clean.

PR #2337, opened 2020-08-15
Partially fixes #566

### LINKED ISSUE #566: Deleting articles had an odd behavior
Attempted to delete a single article and all the articles disappeared, came back and when I checked the edit menu it said “undo delete feed.” I had no intention of deleting a feed.

stuartbreckenridge:
- Adds a preference to show alerts to confirm deletion of feeds (default is true) and this is configurable in Settings / Preferences
- Supports single (iOS/macOS) and multiple selection (macOS) for deletion

Known Issues:
- Until folders are selectable (future beta, hopefully), selecting them for deletion is disabled.
- The macOS sidebar doesn't always refresh after a deletion.

Delete confirms


vincode-io:
The iOS setting doesn't need its own section.  It can be included with FEEDS and just have the text "Confirm when deleting".

Issue #2444, opened 2020-09-21
Unauthorized


punkish:
I don't know what else to say. This is on OS X 10.15.6 with NNW 5.1. Clicking on the refresh button causes the following to popup

![image](https://user-images.githubusercontent.com/349242/93768299-d29a9100-fc36-11ea-9e2d-e344ae70a73c.jpeg)

brentsimmons:
Are you using a syncing system? Which one? Has it worked previously?

punkish:
Yes, I am using Feedly. It has been a while since I launched the Mac version of NNW. All my feeds showed up, and the previously read articles, starred, etc. all show up. But when I click the refresh button, I get that mysterious dialog. I will delete my account (from NNW) and re-add it, and report back if the error persists.

punkish:
ooooh, even weirder. So, I deleted my Feedly account, (was warned that it can't be undone), and then re-added it. NNW asked my permission to use Feedly login to add the account. I clicked yes, entered my login creds, and NNW went to the background (Safari popped back to the foreground). That was odd. I clicked on the NNW icon on the menubar to bring it to the front, and NNW asked me to add the Feedly account again. For kicks, I repeated the cycle. And then, NNW asked me again. Now I have three copies of my Feedly account added to NNW, and it still is waiting for me to say yes a fourth time.

![image](https://user-images.githubusercontent.com/349242/93775290-dbdc2b80-fc3f-11ea-80af-c595a46b9de5.jpeg)

punkish:
**Update:** Fwiw, deleted the two extra Feedly accounts and pressed the [ Refresh ] button. The mysterious "Unauthorized" dialog did not popup, and NNW refreshed the feeds successfully. So, two steps back, but one step forward :)

blackketter:
I had a similar issue when updating to 5.1.  Despite being relatively careful I somehow added two Feedly accounts to NNW.  Deleted both and tried to add one again, but ended up with two. My recollection is that the adding and deleting was slow or unresponsive. At the time I chalked it up to me being impatient and clicking add or delete twice.  I did not see the Unauthorized error though.

vincode-io:
Do we think #2448 fixed this issue?  It looks to me like we were trying to sync multiple Feedly accounts of which only one could be valid at a time.

kielgillard:
[...]

PR #2453, opened 2020-09-23
Clarify the UX of authorizing a Feedly account by directing users…

kielgillard:
…towards their default web browser #2444

Please check the `NSAlert` begin/end sheet dance. My AppKit is rusty.

image


vincode-io:
The alert code looks ok to me.  We just need to change the pull request so that it is against `mac-release`.

kielgillard:
Replaced with https://github.com/Ranchero-Software/NetNewsWire/pull/2468 because the base couldn't be easily changed from `mac-candidate` to `mac-release`.

Issue #778, opened 2019-06-20
TheOldReader: Syncing


vincode-io:
Service: https://theoldreader.com

API Documentation: https://github.com/theoldreader/api

TheOldReader uses the Google Reader API. We have this implemented as ReaderAPI and it is currently being used with FreshRSS.

Ideally we would make the ReaderAPI implementation work with minimal changes and minimal checking of the AccountType. We don't want to break FreshRSS. If things start getting to convoluted in the ReaderAPI we should copy it to its own implementation for TheOldReader.

y0y0z:
@brentsimmons: could you please share your time estimation for releasing version 5.2 alpha?

brentsimmons:
Any time estimation I might come up with would be a complete wild guess, and would probably be wrong in some direction by several months. Sorry about that!

vincode-io:
This has been implemented and is scheduled to be part of the 6.0 NetNewsWire release.  Watch the NetNewsWire News blog because we will be announcing test versions of NetNewsWire 6.0 and really need some help testing this sync service.

Issue #777, opened 2019-06-20
BazQux: Syncing


vincode-io:
Service: https://bazqux.com

API Documentation: https://github.com/bazqux/bazqux-api

Bazqux uses the Google Reader API.  We have this implemented as ReaderAPI and it is currently being used with FreshRSS.

Ideally we would make the ReaderAPI implementation work with minimal changes and minimal checking of the AccountType.  We don't want to break FreshRSS.  If things start getting to convoluted in the ReaderAPI we should copy it to its own implementation for Bazqux.

WanderingStar:
I'm happy to test this.

aredridel:
Same.

rodalpho:
Really surprised NNW doesn't support Bazqux, as it uses the Google Reader API. This is the first RSS app I've run across that doesn't support it, and I've tried them _all._

brentsimmons:
It’s because this is a new app, despite the version number, and we prioritized Feedbin and Feedly first.

rodalpho:
Ahh that explains it. Most RSS apps have been around for a very long time now.

odinian:
> Ahh that explains it. Most RSS apps have been around for a very long time now, in one form or another.

Well, NNW has been around, in one form or another, since 2002. :-)

I too would like BazQux support.

brentsimmons:
It’s a unique situation. While NetNewsWire has been around since 2002, NetNewsWire 5 is nevertheless a brand-new app.

It started out life as an app named Evergreen. We changed the name to NetNewsWire when we got the name back from Black Pixel.

But, under the hood, it really is a 1.0 app!

I know that’s confusing, but that’s the situation.

rodalpho:
No problem, perfectly understandable. Thanks for resurrecting the app, I look forward to trying it out when Bazqux is supported!

vincode-io:
This has been implemented and is scheduled to be part of the 6.0 NetNewsWire release.  Watch the NetNewsWire News blog because we will be announcing test versions of NetNewsWire 6.0 and really need some help testing this sync service.

Issue #416, opened 2018-08-31
Inoreader support


venkatasg:
http://www.inoreader.com/developers/

I love the app so far, and I'd love to see it have support for inoreader, since that's the service I use.

jbeker:
Per conversation with Brent and Maurice, I'll start researching.

venkatasg:
Any way I can help? I've never worked on this sort of stuff but I think I can be of some help :)

jbeker:
@vincode-io I have been looking into this and the basic API is similar to Google Readers, but the big difference is authentication, which is OAuth2. This will require a 2-phase UI to accomplish that will probably need to be much different in iOS vs. Mac. I saw you had put a placeholder OAuth2 authentication type in; did you have thoughts on how this could be implemented?

vincode-io:
I haven't given it any thought.  I know that we will have to do OAuth2 for Feedly, so it's on my list of things to do.

vincode-io:
This has been implemented and is scheduled to be part of the 6.0 NetNewsWire release.  Watch the NetNewsWire News blog because we will be announcing test versions of NetNewsWire 6.0 and really need some help testing this sync service.

PR #2527, opened 2020-10-26
Additional iOS account types

stuartbreckenridge:
Adds additional scene for Inoreader, The Old Reader, Bazqux, and FreshRSS on iOS.


ushuz:
Let's connect some related issues for future reference: #416 #777 #778

PR #2530, opened 2020-10-28
Prototype Refresh

stuartbreckenridge:
• Add accounts refreshed to look a bit more like Mail.


stuartbreckenridge:
Screenshot 2020-10-28 at 10 23 42 PM

vincode-io:
@stuartbreckenridge It looks like there is a conflict you need to resolve.  Also rather than commenting out lines, let's just delete them.  We can always get them back using Git if we need to.

stuartbreckenridge:
@vincode-io Fixed.

PR #2535, opened 2020-10-30
Add account cleanup

stuartbreckenridge:
• Minor tweaks to the iCloud icon
• Consistent footers on add account sheet
• Consistent alignment
• Old functionality removed when the accounts tableview has no selection

Screenshot 2020-10-30 at 9 43 41 AM

Screenshot 2020-10-30 at 9 44 06 AM


kevinm6:
I can’t wait for the new iCloud Sync between devices in this new app version!
After the release, it will be the moment when I get rid of Feedly sync!

PR #2547, opened 2020-11-02
Explainer text for accounts and extensions

stuartbreckenridge:
Explainer text will display when no account/extension is selected.

Screenshot 2020-11-02 at 11 06 16 AM
Screenshot 2020-11-02 at 11 06 21 AM

### DIFF
--- a/Mac/Preferences/Accounts/AccountsPreferencesViewController.swift
+++ b/Mac/Preferences/Accounts/AccountsPreferencesViewController.swift
@@ -226,7 +226,7 @@ private extension AccountsPreferencesViewController {
 	
 	func showController(_ controller: NSViewController) {
 		hideController()
-		
+	
 		addChild(controller)
 		controller.view.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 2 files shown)


vincode-io:
I like it, but the text looks too low.  I'm sure it is centered vertically, but I think the white table view is throwing it off.  Maybe if we centered on the table view?

stuartbreckenridge:
I'm not sure I _see_ the alignment issue. Maybe it looks off because of the +/- buttons?

vincode-io:
I think you are saying the same thing I was trying to say. The left side makes the text look lower in the window than it actually is. I think we should try centering it between the top of the white area and the top of the buttons.

stuartbreckenridge:
I think I'd go for the middle option. 

Offsets

vincode-io:
I agree.  The middle one looks best.

stuartbreckenridge:
Fixed in 3580739 using `NSConstraints` on the `NSHostingController` instead of `.offset` on the SwiftUI view.

Issue #2542, opened 2020-11-01
NewsBlur: Can't rename Feeds or Folders to something with a plus sign (+) in them


vincode-io:
Trying to rename a folder to something like "News + Weather" causes all kinds of problems and renaming a feed to something like "Tools + Fun" drops the plus sign.

PR #2564, opened 2020-11-08
NewsBlur: Fix renaming feed/folder not working with plus sign in name

quanganhdo:
Resolves https://github.com/Ranchero-Software/NetNewsWire/issues/2542

The bug was caused by `+` not being properly handled. 

From https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems:

> RFC 3986 specifies which characters must be percent-encoded in the query component of a URL, but not how those characters should be interpreted.

> One notable example of potential interoperability problems is how the plus sign (+) character is handled:

> According to RFC 3986, the plus sign is a valid character within a query, and doesn't need to be percent-encoded. However, according to the W3C recommendations for URI addressing, the plus sign is reserved as shorthand notation for a space within a query string (for example, ?greeting=hello+world).

### DIFF
--- a/Account/Sources/Account/NewsBlur/Models/NewsBlurFeedChange.swift
+++ b/Account/Sources/Account/NewsBlur/Models/NewsBlurFeedChange.swift
@@ -44,6 +44,10 @@ extension NewsBlurFeedChange: NewsBlurDataConvertible {
 			}
 		}()
 
-		return postData.percentEncodedQuery?.data(using: .utf8)
+		// `+` is a valid character in query component as per RFC 3986 (https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems)
+		// workaround:
+		// - http://www.openradar.me/24076063
+		// - https://stackoverflow.com/a/37314144
+		return postData.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B").replacingOccurrences(of: "%20", with: "+").data(using: .utf8)
 	}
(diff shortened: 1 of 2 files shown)


Account/Sources/Account/NewsBlur/Models/NewsBlurFeedChange.swift
```diff
@@ -44,6 +44,10 @@ extension NewsBlurFeedChange: NewsBlurDataConvertible {
 			}
 		}()
 
-		return postData.percentEncodedQuery?.data(using: .utf8)
+		// `+` is a valid character in query component as per RFC 3986 (https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems)
+		// workaround:
+		// - http://www.openradar.me/24076063
+		// - https://stackoverflow.com/a/37314144
+		return postData.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B").replacingOccurrences(of: "%20", with: "+").data(using: .utf8)
```
quanganhdo, line 51: This may make more sense as an extension to `URLComponents` (in `RSWeb`?) seeing that there's a similar issue for FeedWrangler.

vincode-io:
@quanganhdo I took your suggestion and added the code to RSWeb and updated the NewsBlur code to use it.  Thanks for doing all that research and coming up with a solution.  I'm closing out this pull request since we handled it differently.

PR #2568, opened 2020-11-12
Handle +’s in the query items

### LINKED ISSUE #2544: FeedWrangler: Can't use plus sign (+) when renaming a feed
The plus sign doesn't sync to the server when used in a name like "Food + Drink" for a feed.

jbennett:
It looks like the FeedWrangler service doesn't handle the %2B correctly, which I've emailed them about, but this update at least has us creating the correct URL for the request.

This private extension might be worth moving into RSWeb.

Fixes #2544

### DIFF
--- a/Account/Sources/Account/FeedWrangler/FeedWranglerAPICaller.swift
+++ b/Account/Sources/Account/FeedWrangler/FeedWranglerAPICaller.swift
@@ -111,13 +111,17 @@ final class FeedWranglerAPICaller: NSObject {
 	}
 	
 	func renameSubscription(feedID: String, newName: String, completion: @escaping (Result<Void, Error>) -> Void) {
-		let url = FeedWranglerConfig.clientURL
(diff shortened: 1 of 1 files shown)


vincode-io:
Rather than crashing when `urlWithEnhancedPercentEncodedQuery` executes, let's call the completion handler with an error, any error, so that the user knows something went wrong even if there isn't anything they can do about it.

jbennett:
This has a static URL/path, so the only way this could fail is if it can't encode the query parameters. Is that something that can even happen? If not, would this be an appropriate place for a force unwrap? Otherwise, I'll throw up an error: "Congrats, you broke the interwebz" :D

vincode-io:
I can't foresee how this could happen either.  Better safe than sorry.  Any generic error message is better than crashing and also better than "Congrats, you broke the interwebz".  lol

Issue #2240, opened 2020-07-09
WidgetKit: Add an NNW iOS14/macOS 11 Widget


stuartbreckenridge:

vincode-io:
Aren't Widgets cross platform?  If we make one, it should probably work for macOS too right?

stuartbreckenridge:
Correct. Amended the title

vincode-io:
I wonder if our lack of remove push notifications is going to make a widget a poor user experience.  I still need to watch the WWDC session on them, so I don't know.

If the NNW isn't running on the Mac, will we still get updates for the widget?  If background sync is being flaky on iOS, will we still get an up to date widget?

stuartbreckenridge:
I haven't done the code-along yet. 

My understanding is the widget can trigger a background download.

stuartbreckenridge:
Are there instructions for adding new targets and making sure they pick up the settings from `SharedXcodeSettings`? 

I'm missing something as, unless I type in my organisation identifier, it just sees `group..NetNewsWire.iOS`.

vincode-io:
There aren't any instructions.  We don't do it often enough to document it.  If you want to open a pull request with what you have, I'll finish up the target configuration for you.

stuartbreckenridge:
I will park this for the time being until we work out a way forward. If we go for a remote notification route then I think the work on the branch will be effective as it just writes out a selection of the latest data to JSON which the widget can use.

stuartbreckenridge:
I've picked this up again...it's more useful with remote notifications (which I have running), but there is still utility without them and just background refresh (e.g. for feeds that don't have notifications enabled).

![iVBORw0KGgoAAAANSUhEUgAABwgAAA1ICAYAAACDu8CkAAAACXBIWXMAAAsTAAALEwEAmpwYAAALMWlUWHRYTUw6Y29tLmFkb2Jl-2](https://user-images.githubusercontent.com/7046652/99251251-fb5e9180-2847-11eb-9ce3-8c00f48770fb.PNG)

jean-io:
@stuartbreckenridge this looks great, can't wait to have it my iPhone 🤩

PR #2575, opened 2020-11-18
Widget Bundle

stuartbreckenridge:
For #2240, this PR adds seven widgets:

- A small Smart Feeds summary widget
- 3x medium widgets for Today, Unread, and Starred
- 3x large widgets—larger versions of the above
  - The medium/large widgets support deep linking and navigation to the tapped article, otherwise they will go to appropriate table view
  - They show up to three or eight articles and then an additional count at the bottom 

Note: in order to get this build I had to disable the NoBS in the project file script. I couldn't work out how to get around it.

Image


iOS/SceneDelegate.swift
```diff
@@ -89,6 +92,96 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+				if context.url.absoluteString.contains(WidgetDeepLink.unread.url.absoluteString) {
+					guard let comps = URLComponents(string: context.url.absoluteString ) else { return  }
+					let id = comps.queryItems?.first(where: { $0.name == "id" })?.value
+					if id != nil {
+						if AccountManager.shared.isSuspended {
+							AccountManager.shared.resumeAll()
+						}
+						let articles = try! AccountManager.shared.fetchArticles(.unread)
+						let article = articles.filter({ $0.articleID == id }).first
+
```
vincode-io, line 110: You don't need to fetch all the articles just to tell the coordinator which to select.  It can select based on articleID.  Let's expose the `selectArticleInCurrentFeed` function in `SceneCoordinator` and use that instead.  More on that coming...

iOS/SceneDelegate.swift
```diff
@@ -89,6 +92,96 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+							AccountManager.shared.resumeAll()
+						}
+						let articles = try! AccountManager.shared.fetchArticles(.unread)
+						let article = articles.filter({ $0.articleID == id }).first
+						
+						if article != nil {
+							self.coordinator.selectAllUnreadFeed()
+							DispatchQueue.main.asyncAfter(deadline: .now() + 0.5, execute: {
+								self.coordinator.selectArticle(article!, animations: [.navigation, .scroll])
+							})
```
vincode-io, line 115: Instead of doing an `asyncAfter` we can change `selectAllUnreadFeed` and the other smart feed selects to have completion handlers.  The `selectFeed` function they all rely on has a completion handler on it already.  You would just have to pass the new one to it.  You should then be able to write something like this:

```swift
self.coordinator.selectAllUnreadFeed() {
   self.coordinator.selectArticleInCurrentFeed(articleID)
}
```

stuartbreckenridge: This is what is looks like now:

```swift
if id != nil {
	if AccountManager.shared.isSuspended {
		AccountManager.shared.resumeAll()
	}
	self.coordinator.selectAllUnreadFeed(completion: {
		self.coordinator.selectArticleInCurrentFeed(id!)
	})
} else {
	self.coordinator.selectAllUnreadFeed()
}
```

However, it still needs a dispatch after delay because it's trying to select the article too quickly. What happens at the moment is that it select the unread feed and then just sits there. With the delay it then moves to the article after the initial unread animation has completed.

iOS/SceneDelegate.swift
stuartbreckenridge: Fixed.

stuartbreckenridge:
Fixed in 361ca85.

jean-io:
Hello there, when can we expect having widgets within the app from the appstore?
Thank

PR #2594, opened 2020-11-19
changes the behaviour of the today widget

stuartbreckenridge:
filters out read articles and changes strings to highlight what's on show

### DIFF
--- a/Shared/Widget/WidgetDataEncoder.swift
+++ b/Shared/Widget/WidgetDataEncoder.swift
@@ -50,7 +50,7 @@ struct WidgetDataEncoder {
 			}
 			
 			// Today Articles
-			let todayArticles = try SmartFeedsController.shared.todayFeed.fetchArticles().sorted(by: { $0.datePublished ?? .distantPast > $1.datePublished  ?? .distantPast  })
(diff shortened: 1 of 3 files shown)


Shared/Widget/WidgetDataEncoder.swift
```diff
@@ -50,7 +50,7 @@ struct WidgetDataEncoder {
 			}
 			
 			// Today Articles
-			let todayArticles = try SmartFeedsController.shared.todayFeed.fetchArticles().sorted(by: { $0.datePublished ?? .distantPast > $1.datePublished  ?? .distantPast  })
+			let todayArticles = try SmartFeedsController.shared.todayFeed.fetchArticles().sorted(by: { $0.datePublished ?? .distantPast > $1.datePublished  ?? .distantPast  }).filter({ $0.status.read == false })
 			var today = [LatestArticle]()
```
vincode-io, line 54: I think you want to use `fetchUnreadArticles` here instead of manually filtering them.  Same on line 67.

stuartbreckenridge: Yup!

PR #2601, opened 2020-11-21
Widget Localized

stuartbreckenridge:
This commit uses [SwiftGen](https://github.com/SwiftGen/SwiftGen) to generate typesafe localis(z)ed strings for use in the Widget. 

```bash
swiftgen run strings -t structured-swift5 "$PROJECT_DIR/Widget/Resources/en.lproj/Localizable.strings" "$PROJECT_DIR/Widget/Resources/Localizable.stringsdict" --output "$PROJECT_DIR/Widget/Resources/Localized.swift";
```

The run script is commented out so the app continues to build normally.

It tidies up certain areas of code very nicely:

For example:

```swift
let formatString = NSLocalizedString("UnreadCount",  comment: "Unread Count Format")
let str = String.localizedStringWithFormat(formatString, UInt(count))
```

is now:

```swift
let str = L10n.unreadCount(count)
```

I think this forms a reasonable PoC for how NNW 7.0 _could_ be localised. The process would be:

1. Add all non-localised strings to `Localizable.strings` or `Localizable.stringsdict`
2. Add SwiftGen to the build phases 
3. Use the `L10n.stringName` output in SwiftUI views instead of `NSLocalizedString`s.


vincode-io:
It looks promising.

How does the unreadCount get defined?  I couldn't find an input for it to be generated anywhere.

stuartbreckenridge:
Today, Unread, and Starred counts—any plurals—are pulled from `Localizable.stringsdict`. 

Screenshot 2020-11-21 at 09 55 30

```swift
internal static func unreadCount(_ p1: Int) -> String {
    return L10n.tr("Localizable", "UnreadCount", p1)
}
```

vincode-io:
I like it.  It is certainly better than using NSLocalizedStrings everywhere.

stuartbreckenridge:
It's probably a few days work to get the UIKit/AppKit apps to scratch if we felt there was a benefit in doing so for NNW 6.

brentsimmons:
I don’t mind doing this for the widget, but I want to do more research before deciding how it should be done for the bulk of the app.

Issue #2617, opened 2020-11-25
Newsblur: Unread articles are downloaded even if we have already downloaded them


vincode-io:
During a Newsblur sync we always download all unread articles even if we already have a copy of them in the local database.  We should only download unread article id's (hashes) and then download any missing unread articles.

PR #2620, opened 2020-11-30
Changes to NewsBlur refreshAll process

quanganhdo:
Resolves https://github.com/Ranchero-Software/NetNewsWire/issues/2617
Related conversation: https://netnewswire.slack.com/archives/CCGTPJDPD/p1605910165150500

WIP.


vincode-io:
@quanganhdo Do we need to leave this as a WIP for some reason?

quanganhdo:
The missing part is related to the issue you raised in the Slack thread (the current code not being able to download read articles).

vincode-io:
I don't think we will be able to handle it.  It just looks like a deficiency in the NewsBlur API.  All the syncing services are a little bit different.  This is just something we'll have to live with.  It won't impact our unread counts or starred articles, so it isn't the end of the world.

vincode-io:
It is ok for us to move forward without being able to download unread articles, so let's just go ahead with this pull request.

PR #2686, opened 2020-12-23
Richer notifications

stuartbreckenridge:
On iOS this adds feed icons to the notification content along with three actions for Starred, Read, Open. On macOS, only the actions are added. 

@vincode-io Is there a better way to grab the url of the cached favicon?

In the `UserNotificationManger`, I'm grabbing the image and adding it to the caches directory, and then letting notifications pull it from there.

```swift
func thumbnailAttachment(for article: Article, webFeed: WebFeed) -> UNNotificationAttachment? {
		if let image = article.iconImage() {
		let fm = FileManager.default
		var path = fm.urls(for: .cachesDirectory, in: .userDomainMask)[0]
		#if os(macOS)
		path.appendPathComponent(webFeed.webFeedID + "_smallIcon.tiff")
		#else
		path.appendPathComponent(webFeed.webFeedID + "_smallIcon.png")
		#endif
		fm.createFile(atPath: path.path, contents: image.image.dataRepresentation()!, attributes: nil)
			
		let thumbnail = try? UNNotificationAttachment(identifier: webFeed.webFeedID, url: path, options: nil)
		return thumbnail
	}
	return nil
}
```

Comparison

### DIFF
--- a/Mac/AppDelegate.swift
+++ b/Mac/AppDelegate.swift
@@ -455,7 +455,17 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSUserInterfaceValidations, 
     }
 	
     func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
-		mainWindowController?.handle(response)
+		
+		let userInfo = response.notification.request.content.userInfo
+		
+		switch response.actionIdentifier {
+		case "MARK_AS_READ":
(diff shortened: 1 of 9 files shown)


vincode-io:
I can't think of any better way to get a URL for the Image.  You might want to put the URL determination code in`IconImage` incase we need it again for something else.

vincode-io:
Sorry, I think I merged these in the wrong order.  @stuartbreckenridge Could you remove this merge problem?

stuartbreckenridge:
I think that should be it.

PR #2769, opened 2021-01-29
Text position on widgets

### LINKED ISSUE #2737: Widget: Vertical distance of the count to the bottom of the widget increases when the count < 3
Screenshot 2021-01-17 at 17 11 54
Screenshot 2021-01-17 at 17 12 24

stuartbreckenridge:
Count text is now placed under the article list. 

- This keeps the widget _icon_ leading `VStack` at a consistent size—it was previously changing based on Unread/Starred/Today text.
- It also stops the count text from moving based on the number of articles on display.

Simulator Screen Shot - iPhone 12 Pro - 2021-01-29 at 22 43 18

Fixes #2737


vincode-io:
Should we display any text at all when there are not more items?  I find it confusing that we show "No more starred articles" when there are some showing.

stuartbreckenridge:
Ok. Will remove it.

stuartbreckenridge:
@vincode-io Fixed in 474fbed

vincode-io:
Did you intend to roll back the resolved versions of some packages in that commit?

PR #2790, opened 2021-02-06
: class changed to : AnyObject for protocols

stuartbreckenridge:
`: class` is deprecated in 5.4 but the change to `: AnyObject` can be done now.


brentsimmons:
Has this change been tested on Big Sur and on Catalina?

stuartbreckenridge:
It made zero difference on Big Sur. 

As per [this](https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507/10) discussion, `class` and `AnyObject` are parsed the same way.

Issue #1695, opened 2020-01-25
Search should also search author field


pdokas:
When doing a search in NNW it looks like it searches in the item’s title and description.

I sometimes want to search by author, e.g. to find columns by an author in a newspaper’s opinion section.

I wish search would also include the author field to let this happen.

pdokas:
🎉 I’m so excited, thank you @banaslee and @vincode-io!

vincode-io:
Reopening while we make some minor changes to the implementation.

vincode-io:
This was merged.

pdokas:
@vincode-io It was merged but I don't think it works. Using current TestFlight (6.0 605):

![image](https://user-images.githubusercontent.com/190134/122317006-a19f7200-ced1-11eb-8378-70a108bf1342.png)
![image](https://user-images.githubusercontent.com/190134/122317016-a5cb8f80-ced1-11eb-83a4-558f02ac2430.png)

Searching in the Daring Fireball feed for "Gruber" yields 0 articles.

vincode-io:
That's because it is part of the NNW 6.1 release.

pdokas:
@vincode-io I just pulled the NNW iOS 6.1 TF build and I’m not seeing searches for displayed author names return any articles, e.g. "Gruber" brings up zero DF posts.

vincode-io:
@pdokas I just tried it and it worked for me. You did select the "All Articles" option, right?

vincode-io:
Another thing to keep in mind is that it won't work retroactively. It only works for new articles received after the upgrade to 6.1. Reindexing the whole database would be too CPU intensive, so we decided not to do it.

pdokas:
Ah, got it. I was searching the backlog of articles loaded before installation. Confirmed it works with something received since! Yay 😄

And happy anniversary to this ticket!
<img width="294" alt="image" src="https://user-images.githubusercontent.com/190134/150841744-8701d53a-46f0-45a7-b61d-c9abc4bad9e3.png">

PR #2802, opened 2021-02-10
Adds author name as a searchable field

banaslee:
Fixes https://github.com/Ranchero-Software/NetNewsWire/issues/1695

Users are now able to search for an author name and get as results, articles by authors with that name

Did this by adding an `authors` column to the `search` table and filling it correctly when building the text index
Database also migrates correctly when coming from a previous version by recreating the `search` table and the index

### DIFF
--- a/ArticlesDatabase/Sources/ArticlesDatabase/ArticlesDatabase.swift
+++ b/ArticlesDatabase/Sources/ArticlesDatabase/ArticlesDatabase.swift
@@ -76,6 +76,7 @@ public final class ArticlesDatabase {
 		self.retentionStyle = retentionStyle
 
 		try! queue.runCreateStatements(ArticlesDatabase.tableCreationStatements)
+        try! queue.runCreateStatements(ArticlesDatabase.searchTableCreationStatements)
 		queue.runInDatabase { databaseResult in
 			let database = databaseResult.database!
 			if !self.articlesTable.containsColumn("searchRowID", in: database) {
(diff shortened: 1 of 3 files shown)


vincode-io:
@banaslee Could you change this so that it doesn't do a full reindexing of the search table. @brentsimmons feels it would be better to just index the author field for new and updated articles.

banaslee:
Sure @vincode-io. That should be a very small change. Do you have a suggestion on how to reliably test updated articles case?

vincode-io:
As new articles come in, they should be able to be found in the search function when searching for an authors name. I would start with a new local account and subscribe to the Washington Posts feeds. Then as articles come in, search for them by author name.

banaslee:
Ok @vincode-io, it's more complex than I expected...

See, it's not possible to just add a column to a virtual table like a fts4 table. So, on an upgrade scenario, we need to recreate the table in order to add the column. By recreating it, we lose its contents unless we reindex...

Let me share what I was trying to do:
I was going to delete the code that sets `SearchRowID` to null for all the articles. I was doing this to force a rebuild of the index. Though, what I found in my tests, is if I don't do it, the values in `SearchRowID` become outdated and search stops working.

Any ideas?

I guess the slowness you experienced happens when there's a huge number of articles in the database. Right? As I'm running the app with the default feeds, I'm not running into that problem.

brentsimmons:
Instead of adding a field, how about just appending the author name as text to the body in the search table?

banaslee:
Ah @brentsimmons! Thanks, forgot about that (I had entertained that idea in Slack but dropped it). 

What are your thoughts on `body` now including more than the body?

brentsimmons:
I don’t love it, but it seems like the pragmatic thing here, so it’s fine.

vincode-io:
If we ever decided we wanted to target the author field in a search, we have the option then of reindexing everything. We don't have to do it right now.

banaslee:
@vincode-io I didn't understand your comment: what's the difference between "target the author field in a search" and what we're doing here?

vincode-io:
So that you could search for `author:Simmons` and only get article penned by people with the name Simmons, but not articles where Simmons is mentioned. AppKit and UIKit have ways of turning that `author:` portion into a token with a special user interface.

banaslee:
Got it. If I take @brentsimmons suggestion I'm not sure that will be possible. But is always possible to go back and add an `authors` column like I initially did.
I'm now trying @brentsimmons suggestion 🤞

banaslee:
New PR at https://github.com/Ranchero-Software/NetNewsWire/pull/2804

PR #2813, opened 2021-02-16
Fixes #2781

### LINKED ISSUE #2781: NNW 6: Main Thread Checker Warnings on iPad
Screenshot 2021-02-05 at 06 23 45

stuartbreckenridge:
Updating the toolbar in response to a UserDefault change notification caused the main thread warnings.

### DIFF
--- a/iOS/MasterTimeline/MasterTimelineViewController.swift
+++ b/iOS/MasterTimeline/MasterTimelineViewController.swift
@@ -471,13 +471,15 @@ class MasterTimelineViewController: UITableViewController, UndoableCommandRunner
 	}
 
 	@objc func userDefaultsDidChange(_ note: Notification) {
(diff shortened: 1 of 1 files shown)


brentsimmons:
What code was it that changed a UserDefaults value outside the main thread? That’s not necessarily a bad thing, but it does raise a red flag.

stuartbreckenridge:
I am not convinced that it is NNW code that is changing the user defaults outside the main thread. 

In each one of our `set { }` blocks in `AppDefaults` I added `assert(Thread.isMainThread)` and they were never triggered, but the `assert(Thread.isMainThread)` I added in the function in the PR was triggered. I also tried using `dump()` on the notification itself, but it didn't reveal anything useful.

brentsimmons:
Cool — thanks for checking!

Issue #2774, opened 2021-02-01
Add Context Menus for iOS 14 


stuartbreckenridge:
Add context menus for:
- Add Item
- Mark All as Read

PR #2825, opened 2021-02-23
Menu icons changed

stuartbreckenridge:
Uses more appropriate icons for the menus introduced in #2774


stuartbreckenridge:
The above changes result in this:

Screenshot 2021-02-23 at 21 57 28

vincode-io:
Can we put these in the same order as the Mac menu without the separator? It will keep the two applications more consistent.

vincode-io:
We should probably make this change on `ios-candidate`.

stuartbreckenridge:
Created a separate PR for this #2829 on `ios-candidate`.

PR #2855, opened 2021-03-08
Fixes #2823

### LINKED ISSUE #2823: Article content is in wrong position after window resize due to monitor configuration change
When I unplug my external monitor that has the NNW window on it, the window is resized to fit full screen on my MacBook Pro. 

When the window reappears the article content is offset vertically downward incorrectly. (first pic below)  Resizing the window causes the content to appear in the correct position (second pic)

Screen Shot 2021-02-22 at 9 20 43 PM

Screen Shot 2021-02-22 at 9 21 09 PM

stuartbreckenridge:
Fixes #2823

### DIFF
--- a/Mac/MainWindow/Detail/DetailWebView.swift
+++ b/Mac/MainWindow/Detail/DetailWebView.swift
@@ -54,8 +54,14 @@ final class DetailWebView: WKWebView {
 		super.viewDidEndLiveResize()
 		evaluateJavaScript("document.body.style.overflow = 'visible';", completionHandler: nil)
 		
+		/*
+		On macOS 11, when a user exits full screen
+		or exits zoomed mode by disconnecting an external display
+		the webview's `origin.y` is offset by a sizeable amount.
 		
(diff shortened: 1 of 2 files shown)


Mac/MainWindow/Detail/DetailWebView.swift
```diff
@@ -53,6 +53,19 @@ final class DetailWebView: WKWebView {
 	override func viewDidEndLiveResize() {
 		super.viewDidEndLiveResize()
 		evaluateJavaScript("document.body.style.overflow = 'visible';", completionHandler: nil)
+		
+		
+		// On macOS 11, when a user exits full screen or exits zoomed mode (full screen with menu bar showing) by disconnecting an external display, the webview's origin.y is offset by a sizeable amount. This function adjusts the height of the window height by -1pt/+1pt which puts the webview back in the correct place. This is an issue with SwiftUI and AppKit.
```
vincode-io, line 58: This is a really long comment. I like the details, but could we put it on more than one line?

Issue #2371, opened 2020-08-21
Footnotes Not Working With Articles From Feedly


Nonameentered:
Normal clicks do not do anything. 3D/Force/Haptic touch is able to open the link.
Footnote example here: https://daringfireball.net/2020/08/applespeak_to_english_xbox_game_pass.

ybbond:
Seconded this

brehaut:
I can reproduce this on macOS NNW as well as in iOS. 

It looks like feedly is rewriting the document in some fashion that is throwing off the newsfoot script. Looking in the developer console, whenever the link inside the `.newsfoot-footnote-container` element is clicked I’m seeing a log for

`TypeError: null is not an object (evaluating 'document.querySelector(`[id='${targetId}']`).innerHTML') newsfoot.js: 147`

Looking at the DOM for the footnotes it appears feedly is stripping off the id's from the footnotes themselves, making the above query fail.

I’m not sure if theres necessarily a good work around here. If we could get a small corpus of examples from sites other than Daring Fireball we could figure out a plan.

At a baseline, it probably pays to check the return value of the `querySelector` call and fallback to browser default behaviour if we can't handle it ourselves.

Secondly, at least in the case of Daring Fireball, we have enough information in the footnote reference link and the footnote's return link to stitch together the content. I'm hesitant to build that special case until we know how other sites markup is affected.

vincode-io:
Here is another blog that uses footnotes that has the same issue.  https://pxlnv.com/feed/json

Here is an example that you can pull up in NNW on a Feedly account.

[Reviewers See Simultaneous Gains in Performance and Battery Life With Apple’s First M1 Macs](https://pxlnv.com/linklog/m1-mac-reviews/)

ybbond:
first, for @brentsimmons , I want to clarify. You added "iOS" label for this issue, but this is also happen for macOS version.

second, is there anyone already working on this issue?
I took a little inspection with NNW's inspect element for DaringFireball, MacStories and some other sites fetched from both Feedbin and Feedly. I also take a look at the current `newsfoot.js`. If there is currently no one tackling this, I might try to modify `newsfoot.js`.

[...]

PR #2867, opened 2021-03-10
#2371 Checks footnote target before overriding browser default behavior

brehaut:
This change just checks the return value of the document query to find the footnote element; if not element is found then the browsers default behaviour (sending you to the link in your browser) is not overridden. 

Note that footnotes come with their own return links as part of the markup. These will also take you to the browser if you click on them as they suffer the same issues with missing attributes as the referring link.

### DIFF
--- a/Shared/Article Rendering/newsfoot.js
+++ b/Shared/Article Rendering/newsfoot.js
@@ -140,12 +140,17 @@
 			if (targetId) break;
 		}
 		if (targetId === undefined) return;
-
+		
+		// Only override the default behaviour when we know we can find the
+		// target element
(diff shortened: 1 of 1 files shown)


Shared/Article Rendering/newsfoot.js
```diff
@@ -140,12 +140,17 @@
 			if (targetId) break;
 		}
 		if (targetId === undefined) return;
-
+		
+		// Only override the default behaviour when we know we can find the
+		// target element
+		const targetElement = document.querySelector(`[id='${targetId}']`);
```
j-f1, line 146:
```suggestion
		const targetElement = document.getElementById(targetId);
```

Issue #2875, opened 2021-03-12
Feedly not working in NetNewsWire 6 beta


rhx:
After installing NetNewsWire 6 under Big Sur, my Feedly account stopped refreshing, giving me a "Not authorized" error instead.

Removing the account in Preferences/Accounts, then trying to re-add it gives me a browser pop-up window that lets me log in fine.  But after the browser pop-up closes, I get:

The operation couldn’t be completed. (Secrets.CredentialsError error 0.)

The account shows up, but every time I try to refresh it, I get "Please add the Feedly account again."  When I do that, I get the same error all over again.

rhx:
After digging a little bit, it appears that this is an issue if there is an older keychain entry for Feedly (from the date this appears to have come from when I set up NetNewsWire 5).

Deleting all Feedly entries in the KeyChain, then removing and re-adding the Feedly account fixes the problem.  It's okay if you want to close the issue, but it might be a worth to perhaps change the wording of "Please add the Feedly account again."  to something like
```
Please add the Feedly account again.
If this problem persists, open the KeyChain app, delete all feedly.com entries, then try again.
```

rhx:
I have created Pull request #2876 to fix this issue.

kielgillard:
Strange, this has been working fine for me. Did you rename the NetNewsWire app by any chance?

rhx:
Yes, I have both versions 5 and 6 beta installed, so I renamed 6.0b1 to `NetNewsWire 6.app` prior to moving it to `/Applictions` so it can co-exist with version 5.  Interestingly, after removing the KeyChain entries and re-adding the Feedly account (from NetNewsWire 6), this now works fine for both version 5 and version 6.

kielgillard:
Interesting. What happens if you both move NetNewsWire 5 to a different folder and return the name of "NetNewsWire 6" to simply "NetNewsWire"? Do you eventually get keychain errors again?

rhx:
Renaming NetNewsWire 6 back to NetNewWire doesn't seem to make any difference: it still works.  Hard to speculate what the reason for failure was.  Maybe (given the old dates of the original KeyChain entries), NetNewsWire 6 had permission to read the old (NetNewsWire 5) token/password, but not to change it?  The strange thing is that this happened on both my Laptop and my Desktop (which is not surprising, given the entries are in the login keychain, not the iCloud keychain).

[...]

PR #2876, opened 2021-03-12
Provide additional help to users with persistent Feedly authorization issues

rhx:
Feedly account authorization can get stuck when older KeyChain entries (e.g. from NetNewsWire 5) interfere with proper authorization, even when removing and re-adding a Feedly account.  This change gives the user more information, i.e. to delete `feedly.com` entries from their KeyChain if they encounter a persistent problem.

This closes issue #2875

### DIFF
--- a/Account/Sources/Account/Feedly/FeedlyAccountDelegateError.swift
+++ b/Account/Sources/Account/Feedly/FeedlyAccountDelegateError.swift
@@ -23,7 +23,7 @@ enum FeedlyAccountDelegateError: LocalizedError {
 	var errorDescription: String? {
 		switch self {
 		case .notLoggedIn:
-			return NSLocalizedString("Please add the Feedly account again.", comment: "Feedly – Credentials not found.")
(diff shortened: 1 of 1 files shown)


Account/Sources/Account/Feedly/FeedlyAccountDelegateError.swift
```diff
@@ -23,7 +23,7 @@ enum FeedlyAccountDelegateError: LocalizedError {
 	var errorDescription: String? {
 		switch self {
 		case .notLoggedIn:
-			return NSLocalizedString("Please add the Feedly account again.", comment: "Feedly – Credentials not found.")
+			return NSLocalizedString("Please add the Feedly account again. If this problem persists, open the KeyChain app, delete all feedly.com entries, then try again.", comment: "Feedly – Credentials not found.")
```
j-f1, line 26:
```suggestion
			return NSLocalizedString("Please add the Feedly account again. If this problem persists, open Keychain Access and delete all feedly.com entries, then try again.", comment: "Feedly – Credentials not found.")
```

rhx:
@j-f1 I have applied your recommended changes

PR #2956, opened 2021-04-01
Add Account Sheet

### LINKED ISSUE #2955: Add new account dialog too tall (covered by Dock)
The popover dialog that appears after clicking the "+" button in Preferences > Accounts seems too tall, as the bottom runs under the Dock at the default resolution on a retina MacBook Pro 13" (1600px tall). This is exacerbated by the fact that since it's a window-modal popover, it can't be resized or dragged out of the way to use.

NetNewsWire Version 6.0 (6026) on macOS Big Sur (11.2.3), MacBook Pro (Retina, 13-inch, Mid 2014).

nnw-add-account

stuartbreckenridge:
The web accounts are now chunked into two columns of upto 4 accounts. Fixes #2955.


vincode-io:
We already have `chunked` implemented in `Array+RSCore`. It looks like you also included FeedWrangler in the new list, but it isn't ready to be added yet.

Issue #2976, opened 2021-04-09
Disable Reader View for Feed Provider sourced WebFeeds


stuartbreckenridge:
Disabling Reader View for Feed Provider feeds will stop the Twitter JavaScript warning message from appearing.

In `WebFeed`: 

- [x] Add a property to determine if `WebFeed` is a `FeedProvider`
- When the above property is `true`, the below behaviours should be observed: 
- [x] `isArticleExtractorAlwaysOn` should always return `false` and cannot be set to `true`

On macOS:

- [x] Remove Reader View from Sidebar Context Menu
- [x] Disable Reader View in Feed Inspector
- [x] Disable Reader View in Toolbar when first selected timeline item is a Feed Provider article

On iOS

- [x] Disable Reader View in Feed Inspector
- [x] Disable Reader View in Toolbar when selected article is a Feed Provider article

vincode-io:
Looks like we forgot to close this one.

PR #2977, opened 2021-04-09
Disable reader view

stuartbreckenridge:
Partial work on #2976


Account/Sources/Account/WebFeed.swift
```diff
@@ -220,6 +225,16 @@ public final class WebFeed: Feed, Renamable, Hashable {
 			postUnreadCountDidChangeNotification()
 		}
 	}
+    
+    // MARK: - Feed Provider
+    public var isFeedProvider: Bool {
+        get {            
+            if FeedProviderManager.shared.best(for: URLComponents(url: URL(string: url)!, resolvingAgainstBaseURL: false)!) == nil {
```
vincode-io, line 232: I don't think we should be force unwrapping here. A bad URL, however unlikely, would cause crashes. Let's return false if it is a bad URL.

Mac/MainWindow/MainWindowController.swift
```diff
@@ -1069,6 +1070,15 @@ private extension MainWindowController {
...
 			}
+			
+			if let webfeed = currentTimelineViewController?.selectedArticles.first?.webFeed {
+				if webfeed.isFeedProvider {
+					toolbarButton.isEnabled = false
+					return false
+				} else {
+					toolbarButton.isEnabled = true
+				}
+			}
```
vincode-io, line 1081: Is the toolbar item enabled or disabled when zero or multiple items are selected in the Timeline?

Account/Sources/Account/WebFeed.swift
stuartbreckenridge: Fixed in 0fa30a94.

Mac/MainWindow/MainWindowController.swift
stuartbreckenridge: The extractor is disabled in these scenarios.

PR #3016, opened 2021-04-18
iOS: Feedly API is no longer returning the collection/folder values for deleted collections/folders…

### LINKED ISSUE #3015: Feedly: Failure parsing response for successfully deleting a folder
From @j-f1 on [Slack](https://netnewswire.slack.com/archives/C017LQCQ3LG/p1618612885070000):
> Deleting a folder in a Feedly account (which contains feeds) gives me a “Data could not be read because it is missing” error
> If I restart the app the folder is gone, however.

Similar issue to #2897. Looks like Feedly changed their response for deleting both collections (folders) and feeds.

kielgillard:
…so make the response (which we ignore anyway), optional. Fixes #3015 for iOS.


kielgillard:
Thanks again @j-f1

brentsimmons (approved):

PR #3073, opened 2021-04-25
Improve accessibility action labels for MasterTimelineCell in iOS

### LINKED ISSUE #3072: Improve accessibility action labels for MasterTimelineCell in iOS
In NetNewsWire for iOS, MasterTimelineCells correctly implement accessibility action labels for the functions that can be accessed via swift actions. One of these actions is to mark the article as either read or unread, depending on the current status of the article. The current labels for these actions are "Read" and "Unread". However, VoiceOver voices "Read" to rhyme with _need_ not _bread_. This is actively misleading for a VoiceOver user, because the action "read" (rhymes with _need_) would be expected to open the article for reading... which is after all the primary function of NetNewsWire. Instead, this action marks the article as _having been read_.

Proposed change: rename these actions to "Mark as read" and "Mark as unread". VoiceOver labels should be as short as possible _while being clear_. The current labels are _too short_ to be clear. "Read" turns out to be actively misleading, due to the VoiceOver pronunciation of this ambiguous word, and "Unread" is just less clear as an action than "Mark as unread".

While this change could be trivially made in the code, I propose to make a pull request that includes unit testing to ensure that there is not a future regression in this fix—since it would be easy for a person in the future to unwittingly think that changing these action labels back to "Read" and "Unread" was fine.

A screen recording demonstrates this issue. See the bullet points that follow for further explanation: 

https://user-images.githubusercontent.com/376937/115975078-dbc95300-a5b5-11eb-83fc-89f5c02ea578.mp4

The swipe actions taken by the user in the video:
0) VoiceOver was activated via the usual accessibility settings/shortcuts (before this recording starts)

1) Swipe to the right a series of times, each time stepping through one interface element, until the first tableview cell is selected

2) On reaching the cell, VoiceOver reads out the defined cell content, ending with "Actions available", which indicates there are actions the user can access via this cell.

3) Swipe down a series of time, each one voicing one action available.

4) On the action "Read", double tap on the screen to activate that action.

5) The cell is marked as Read, and the cell contents are voiced again.  (Somewhat confusingly, after initiating this action VoiceOver then reads out the cell VoiceOver content again, including an article preview from the cell again—which is longer than the text displayed in writing on the cell and thus even potentially gives the initial impression that the article _is_ being read out.

babbage:
Fixes #3072, by changing the accessibility labels for MasterTimelineCells to “Mark as Read” and “Mark as Unread”. 

Includes unit tests to protect against future regression. The current iOS app has fairly tight coupling of the SceneCoordinator and MasterTimelineViewController. [EDIT: See later discussion implementing an alternative from the public modifier discussed here:] Additionally, the SceneCoordinator method ```replaceArticles(with unsortedArticles: Set<Article>, animated: Bool)``` had to be marked as public in order to enable injection of test data into the MasterTimelineViewController. This prompts a warning:
>  'public' modifier conflicts with extension's default access of 'private’

Interested in feedback/discussion. My suggestion would be that the tightly coupled nature of these classes should be revised to use dependency injection, in order to allow testing without such work-arounds. However, it is still better to have unit tests for these issues and it would not make sense to delay this minor accessibility feature for what could be quite a significant refactoring. I note that these are the first unit tests for the NetNewsWire iOS target. Here’s to many more. :)

### DIFF
--- a/iOS/MasterTimeline/MasterTimelineViewController.swift
+++ b/iOS/MasterTimeline/MasterTimelineViewController.swift
@@ -243,8 +243,8 @@ class MasterTimelineViewController: UITableViewController, UndoableCommandRunner
 
 		// Set up the read action
 		let readTitle = article.status.read ?
-			NSLocalizedString("Unread", comment: "Unread") :
-			NSLocalizedString("Read", comment: "Read")
+			NSLocalizedString("Mark as Unread", comment: "Mark as Unread") :
+			NSLocalizedString("Mark as Read", comment: "Mark as Read")
 		
 		let readAction = UIContextualAction(style: .normal, title: readTitle) { [weak self] (action, view, completion) in
 			self?.coordinator.toggleRead(article)


babbage:
I would be happy to subsequently work on another pull request that provides a proposed solution for loosening the current tight coupling of the current SceneCoordinator and MasterTimelineViewController, and providing a mechanism at init for injection of test data, rather than making replaceArticles(...) public. But if you'd prefer, I could also work on that as part of this pull request itself. 

That'd substantially broaden the impact and changes of the pull request, so didn't want to go there initially... here and open to guidance of course.

babbage:
Had a re-think on marking the SceneCoordinator function as public, as that seemed like a bad approach. Clearly the need is for a method to inject test data, since these tests need data so that cells are actually generated in the table. The additional commit implements a MockSceneCoordinator, a subclass of SceneCoordinator with a single function:
`func setTestData(with unsortedArticles: Set<Article>, animated: Bool)`

In order to access the underlying `replaceArticles` method, the MockSceneCoordinator needs to be defined in the same file as the SceneCoordinator—the tidy approach of putting it in a TestUtilities folder associated with the NetNewsWire-iOSTests group is not possible as the protection on this method is `fileprivate`.  But it does at least encapsulate this access in a type clearly associated with testing.

babbage:
It would be good to wrap this MockSceneCoordinator in an `#if TEST` if the project was set up with this as a preprocessor macro. Although the Mac app target references this in a number of places it is always in a context like `#if MAC_APP_STORE || TEST` and unless I'm missing something `TEST` isn't actually implemented in the .xcconfig.

brentsimmons (changes requested):
Would you re-do this with just the functionality change? I want to think some more about how to test this kind of thing — let’s handle that separately.

iOS/MasterTimeline/MasterTimelineViewController.swift
```diff
@@ -263,8 +263,8 @@ class MasterTimelineViewController: UITableViewController, UndoableCommandRunner
 		
 		// Set up the star action
 		let starTitle = article.status.starred ?
-			NSLocalizedString("Unstar", comment: "Unstar") :
-			NSLocalizedString("Star", comment: "Star")
+			NSLocalizedString("Unstar", comment: "Accessibility label for action: Unstar") :
+			NSLocalizedString("Star", comment: "Accessibility label for action: Star")
```
brentsimmons, line 267: We should probably drop the “Accessibility label for action:” part of the comment parameters. These labels are not just for accessibility — they’re used as labels that display for swipe actions.

babbage: Of course, done. My rationale when I did this was that in this specific context, those labels aren't being used for display since the swipe actions used in the iOS app only show an icon—so the comment could clarify how the text would end up being used. But that is dependent on the current implementation of the cells. Anyway, removed of course. :)

babbage:
Updated pull request now contains only the functionality change.

If you would like options or contributions to discussion for how to implement testability of this, and the related classes, happy to contribute, but of course also happy to leave this with you. :)

brentsimmons (approved):

PR #3079, opened 2021-04-26
Handle links with unencoded spaces. Fixes #3069

### LINKED ISSUE #3069: Sharing doesn’t work with specific feed
App Store Connect report:

“For some reason I have one feed t3n (https://t3n.de/tag/gadgets-lifestyle/rss.xml) where I could not use the share function. In the list view the share button is missing entirely and in the article view the share button has no function.”

original

babbage:
Issue #3069 identified that when a feed has links that incorrectly include space characters that have not been percent encoded, a user is unable to share an article and in the swipe actions on an article's MasterTimelineTableViewCell, the share option does not even appear under More. These bugs are a result of guard statements that test if an `Article.preferredLink` string exists and whether a URL can be constructed out of this. When a space character is present in the string, the URL cannot be constructed and the function returns.

This pull request addresses ten such locations in NetNewsWire where the presence of a guard statement would cause UI to not be displayed/an action not proceeded with—that could be fixed by replacing space characters with percent encoding. There are also several other places that guard for the presence of a preferredLink but do not test for the ability to covert to a URL. These remain unchanged.

Two points on style:
• Brent Simmons in Slack directed "Fixing up broken URLs is a good idea — it should be done at the point of use (when passing to the web browser to open)." Hence the changes in ten separate locations, if I have understood this correctly. Another option would be to address this at a single point in ArticleUtilities, either by modifying `preferredLink` itself to correct this issue, or by creating an additional convenience computed var `preferredURL`, which could return an optional URL object if both preferredLink existed and a URL could be generated. This would simplify each of these ten guard statements to `guard let url = article.preferredURL else { return }` The latter would have perhaps been my default preference to implement.

• `.replacingOccurrences(of: " ", with: "%20")` has been directly appended to `preferredLink` in each case, as .replacingOcurrences(of:) returns a non-Optional String. If this was not directly appended to preferredLink, there would have to be two guard statements separated by a non-optional line of code, resulting in more noise than seemed warranted.

### DIFF
--- a/Shared/Extensions/ArticleUtilities.swift
+++ b/Shared/Extensions/ArticleUtilities.swift
@@ -56,6 +56,18 @@ extension Article {
 		return nil
 	}
 	
+	var preferredURL: URL? {
+		guard let link = preferredLink else { return nil }
+		// If required, we replace any space characters to handle malformed links that are otherwise percent
+		// encoded but contain spaces. For performance reasons, only try this if initial URL init fails.
+		if let url = URL(string: link) {
+			return url
+		} else if let url = URL(string: link.replacingOccurrences(of: " ", with: "%20")) {
+			return url
+		}
+		return nil
+	}
+	
 	var body: String? {
 		return contentHTML ?? contentText ?? summary
 	}
--- a/Mac/MainWindow/Timeline/ArticlePasteboardWriter.swift
+++ b/Mac/MainWindow/Timeline/ArticlePasteboardWriter.swift
@@ -38,7 +38,7 @@ extension Article: PasteboardWriterOwner {
 	func writableTypes(for pasteboard: NSPasteboard) -> [NSPasteboard.PasteboardType] {
 		var types = [ArticlePasteboardWriter.articleUTIType]
 
-		if let link = article.preferredLink, let _ = URL(string: link) {
+		if let _ = article.preferredURL {
 			types += [.URL]
 		}
 		types += [.string, .html, ArticlePasteboardWriter.articleUTIInternalType]
--- a/Multiplatform/Shared/Article/ArticleToolbarModifier.swift
+++ b/Multiplatform/Shared/Article/ArticleToolbarModifier.swift
@@ -107,7 +107,7 @@ struct ArticleToolbarModifier: ViewModifier {
 					.disabled(sceneModel.shareButtonState == nil)
 					.help("Share")
 					.sheet(isPresented: $showActivityView) {
-						if let article = sceneModel.selectedArticles.first, let link = article.preferredLink, let url = URL(string: link) {
+						if let article = sceneModel.selectedArticles.first, let url = article.preferredURL {
 							ActivityViewController(title: article.title, url: url)
 						}
 					}
--- a/Multiplatform/Shared/Timeline/TimelineModel.swift
+++ b/Multiplatform/Shared/Timeline/TimelineModel.swift
@@ -249,12 +249,11 @@ class TimelineModel: ObservableObject, UndoableCommandRunner {
 	}
 	
 	func openIndicatedArticleInBrowser(_ article: Article) {
-		guard let link = article.preferredLink else { return }
-		
 		#if os(macOS)
+		guard let link = article.preferredLink else { return }
 		Browser.open(link, invertPreference: NSApp.currentEvent?.modifierFlags.contains(.shift) ?? false)
 		#else
-		guard let url = URL(string: link) else { return }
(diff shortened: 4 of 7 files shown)


brentsimmons (changes requested):
I had assumed there were just two places (at most) we’d need to do this (sending to browser on iOS and on Mac) — but clearly I was wrong!

Let’s go with your other idea: adding a preferredURL property to Article. This will give us a single place to fix up URLs. (We might find other issues besides encoding space characters, and this would give us a single place to do that.)

stuartbreckenridge:
Should we also consider removing existing percent encoding and then adding it back in using `Foundation` methods?

E.g. `let escapedURL = url.removingPercentEncoding?.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)`

```swift
var preferredLink: String? {
        if let url = url, 
           !url.isEmpty,
           let escapedURL = url.removingPercentEncoding?.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),
           URL(string: escapedURL) != nil {
            return escapedURL 
        }
        if let externalURL = externalURL, 
           !externalURL.isEmpty,
           let escapedURL = externalURL.removingPercentEncoding?.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),
           URL(string: escapedURL) != nil {
            return escapedURL
        }
        return nil
}
```

The above has the benefit of checking if a URL can be created in advance of it being returned.

babbage:
> Should we also consider removing existing percent encoding and then adding it back in using `Foundation` methods?

Stuart I started down this exact approach when I was working on the original patch, but it's more complicated than you would think. Valid percent encoding varies by URL component. The [Apple Documentation](https://developer.apple.com/documentation/foundation/nsstring/1411946-addingpercentencoding) states, 

> Entire URL strings cannot be percent-encoded, because each URL component specifies a different set of allowed characters. For example, the query component of a URL allows the “@” character, but that character must be percent-encoded in the password component.

In order to take such an approach, we would have to take the link string, break it down into separate components, de-encode them, then re-encode each of them, then construct the URL out of them. Seems to me there are multiple points at which this could go wrong. What we _do_ know is two things—space characters are invalid in all parts of a URL, and that replacing them with percent encoding _anywhere_ won't make it worse than leaving them there. In some cases it'll "fix" the URL, and in other cases it will still not be able to be parsed to a valid url and we're back where we started.

I'll look into the centralised approach. I too was surprised the issue cropped up in so many places. I am assuming we are probably fixing some additional undocumented bugs that appear when there are malformatted URLs as a result.

babbage:
> I had assumed there were just two places (at most) we’d need to do this (sending to browser on iOS and on Mac) — but clearly I was wrong!

The other interesting thing about this is that my testing seemed to show that even with the malformed URL with a space, sending to browser on Mac worked fine... at least from the controls I tested. I'll have another look but it seemed like NetNewsWire must have been sending the string with the space to the browser, and the browser was making the correction...

babbage:
Changed as discussed. As I thought above, problems with actually opening the article in a browser only affected iOS since on macOS NetNewsWire hands the link string to the browser, not a URL object, and the browser was doing the percent encoding.  However, in a couple of places the same issue did affect macOS, e.g., `writableTypes(for pasteboard: NSPasteboard)`. All fixed here.

Shared/Extensions/ArticleUtilities.swift
```diff
@@ -56,6 +56,14 @@ extension Article {
...
 	}
 	
+	var preferredURL: URL? {
+		// Replace any space characters to handle malformed links that are otherwise percent encoded but contain spaces
+		if let link = preferredLink?.replacingOccurrences(of: " ", with: "%20"), let url = URL(string: link) {
+			return url
+		}
+		return nil
+	}
+
```
brentsimmons, line 66: I’m concerned with performance implications if this is called often from the timeline view.

How about this: try creating a URL from preferredLink first and returning that. If it fails, then replace " " with "%20" and try creating a URL again. This will reduce the performance issue down to just URL initing (almost all of the time).

brentsimmons (changes requested):

Shared/Extensions/ArticleUtilities.swift
babbage: Great idea. And given preferredURL is only being called in places where a URL was already being inited previously, there will be no new performance effect at all where the link is well formed.

babbage: Revised as suggested.

brentsimmons (approved):

PR #3089, opened 2021-04-28
Handle footnote backlinks that lack css classes and ids. Fixes #3084 for iOS

### LINKED ISSUE #3084: Footnote behavior inconsistent/linking remotely
On macOS, this link from sixcolors.com - "Apple execs talk AirTags security, pet tracking" - properly displays an inline footnote when you click on the footnote. However, on iOS in NNW 5.0.5 (56), that link incorrectly opens a web browser window to the URL field of the link, rather than displaying the footnote, scrolling down, or doing nothing.

I have a hard time seeing how following the URL link from the JSON feed is the right behavior for tapping on a footnote link in any case, so this seems like a bug.

JSON feed: https://sixcolors.com/?feed=json

babbage:
Handle footnote backlinks that lack css classes and ids. Fixes #3084 for iOS

### DIFF
--- a/Shared/Article Rendering/newsfoot.js
+++ b/Shared/Article Rendering/newsfoot.js
@@ -157,7 +157,7 @@
     document.addEventListener("click", (ev) =>
     {
 	    if (!(ev.target && ev.target instanceof HTMLAnchorElement)) return;
-        if (!ev.target.matches(".footnotes .reversefootnote, .footnotes .footnoteBackLink, .footnotes .footnote-return")) return;
+        if (!ev.target.matches(".footnotes .reversefootnote, .footnotes .footnoteBackLink, .footnotes .footnote-return, .footnotes a[href^='#']")) return;
 		const id = idFromHash(ev.target);
 		if (!id) return;
 		const fnref = document.getElementById(id);
--- a/Shared/Article Rendering/shared.css
+++ b/Shared/Article Rendering/shared.css
@@ -375,7 +375,8 @@ img[src*="share-buttons"] {
 
 .newsfoot-footnote-popover .reversefootnote,
 .newsfoot-footnote-popover .footnoteBackLink,
-.newsfoot-footnote-popover .footnote-return {
+.newsfoot-footnote-popover .footnote-return,
+.newsfoot-footnote-popover a[href^='#fn'] {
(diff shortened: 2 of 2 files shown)


babbage:
Good patch, bad branch base. Will fix shortly. :)

babbage:
@brehaut patch as discussed in Slack. Learned some interesting new things about wildcard CSS selectors that turned out to work quite well. I think possibly with the wildcard selector, the other specific selectors might actually be redundant—but it didn't seem worth removing them as I wasn't sure I had the test data to confirm this and there seems to be close to zero cost keeping them. Tests well—interested in your feedback that I have hit all required points with this patch.

babbage:
Further thought re the above. Definitely the other specific selectors are not redundant, since the above looks for internal links (that start with "#", with no http/https) but there is nothing stopping a feed provider from putting fully resolved links in as their footnote back links... which the existing code would still I think have picked up for those sites. :)

brehaut:
Change looks good to me. Looking at the markup in the feed it appears that the element + attribute selector is close to as specific as we can make it. Possibly `a[href^='#fn']` would be slightly safer; might be a call for @brentsimmons.

Regarding the selectors in general, my preference has been to be err on the side of being conservative, which has mean’t handling specific cases as we encounter them, rather than going for looser rules. My reasoning for this is that because we are changing default behaviour of the markup, and the code has to run against all kinds of wild markup that comes in from the wild, false negative is preferable to a false positive.

babbage:
@brehaut your comment got me thinking. We would want the click behavior in newsfoot.js to apply to all page-internal links that are within the footnote text at the bottom of the page, because whether they are the return to footnote reference link or a link to another part of the same page, you still want them to scroll to that current page position and not open a separate web view. So the slightly less specific "#" selector is best there I suggest. 

For the shared.css your suggested edit of "#fn" is definitely better and quite important. In shared.css we hide the backlink in the pop-up footnote display, and we _wouldn't_ want to hide any other page-internal links that a person had decided to include in their footnote. So I have updated to now use your more selective err... selector. :)

brehaut:
@babbage Oh yeah good point, that logic seems sound

brentsimmons (approved):

PR #3090, opened 2021-04-28
Handle footnote backlinks that lack css classes and ids. Fixes #3083 for macOS

babbage:
Handle footnote backlinks that lack css classes and ids. Fixes #3083 for iOS


babbage:
Good patch, bad branch base. Will fix shortly. :)

babbage:
Replacement macOS pull request based on correct branch to come shortly. Apologies for noise. Feel free to delete this PR. :)

PR #3091, opened 2021-04-28
Handle footnote backlinks that lack css classes and ids. Fixes #3084 for macOS

### LINKED ISSUE #3043: Link not opening when tapping on title in specific feed
App Store Connect report:

“Source link not opening for articles when tap on title. Tested with: http://www.eldiadevalladolid.com/rss/DVlocal.xml”

### LINKED ISSUE #3084: Footnote behavior inconsistent/linking remotely

babbage:
—macOS application of this same patch submitted for the iOS branch.—

### DIFF
--- a/Shared/Article Rendering/newsfoot.js
+++ b/Shared/Article Rendering/newsfoot.js
@@ -157,7 +157,7 @@
     document.addEventListener("click", (ev) =>
     {
 	    if (!(ev.target && ev.target instanceof HTMLAnchorElement)) return;
-        if (!ev.target.matches(".footnotes .reversefootnote, .footnotes .footnoteBackLink, .footnotes .footnote-return")) return;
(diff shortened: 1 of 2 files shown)


babbage:
@brehaut application of same patch to macOS branch. See my discussion notes on the iOS patch.

brentsimmons (approved):

PR #3095, opened 2021-04-29
Capture fully resolved footnote backlinks. Extends #3084

babbage:
If a footnote backlink has a fully resolved URL, not just a detailed same-page # link, it was not caught by the recent patch. The need for this has now been identified in the RSS feed for SixColors.com. (The original bug report was for the json feed, which has slightly different link formatting.)

### DIFF
--- a/Shared/Article Rendering/newsfoot.js
+++ b/Shared/Article Rendering/newsfoot.js
@@ -157,7 +157,7 @@
     document.addEventListener("click", (ev) =>
     {
 	    if (!(ev.target && ev.target instanceof HTMLAnchorElement)) return;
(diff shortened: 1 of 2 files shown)


babbage:
Exactly what this does:
• In the footnote hover box, it will now hide a link if `#fn` appears anywhere in the link, not just at the start.
• In the footnote section itself, it will continue to capture _all_ links that are page-relative but will also capture fully resolved links (containing a domain name at the start) if they have a page-location-target that contains `#fn` as well. In the wild chance that someone links to an external site, with a page anchor in the url, and that page anchor contains `#fn` then that link will fail to work. I say that's just tough. The much more likely scenario is the one we've encountered here, with a base url on the backlink, that doesn't match the canonical url for the article, and where the link should be treated as an on-page link still. This does that, with zero performance effects.

brentsimmons (approved):

PR #3102, opened 2021-05-01
Add Copy Article URL & Copy External URL Menu Items. Fixes #1285.

### LINKED ISSUE #1285: Feature Request: Copy Plain Text Link to Article
It would be nice to be able to copy just the URL of an article. When you right-click the title you can 'Copy Link' but that copies the title as a hyperlink which doesn't always play nice with other apps like Messages.

robmathers:
The Mac items display in the edit and contextual menus, and on iOS in the contextual menu for an article. The external URL option disables/hides as appropriate if there's no external URL, or if it's the same as the article URL.

**iOS**
 

**Mac**
 
Screen Shot 2021-05-01 at 4 51 11 PM Screen Shot 2021-05-01 at 4 51 07 PM

### DIFF
--- a/iOS/MasterTimeline/MasterTimelineViewController.swift
+++ b/iOS/MasterTimeline/MasterTimelineViewController.swift
@@ -362,6 +362,17 @@ class MasterTimelineViewController: UITableViewController, UndoableCommandRunner
 				menuElements.append(UIMenu(title: "", options: .displayInline, children: secondaryActions))
 			}
 			
+			var copyActions = [UIAction]()
+			if let action = self.copyArticleURLAction(article) {
+				copyActions.append(action)
+			}
+			if let action = self.copyExternalURLAction(article) {
+				copyActions.append(action)
+			}
+			if !copyActions.isEmpty {
+				menuElements.append(UIMenu(title: "", options: .displayInline, children: copyActions))
+			}
+			
 			if let action = self.openInBrowserAction(article) {
 				menuElements.append(UIMenu(title: "", options: .displayInline, children: [action]))
 			}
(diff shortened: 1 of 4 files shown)


iOS/MasterTimeline/MasterTimelineViewController.swift
```diff
@@ -887,6 +898,25 @@ private extension MasterTimelineViewController {
 		}
 		return action
 	}
+	
+	func copyArticleURLAction(_ article: Article) -> UIAction? {
+		guard let preferredLink = article.preferredLink, let url = URL(string: preferredLink) else { return nil }
```
babbage, line 903: This pull request is against main (which is indeed what the contributing guidelines say to do) rather than against the two branches where recent commits have been pulled (ios-candidate and mac-release), and so it hasn't picked up on the addition of article.preferredURL—which definitely would be used here instead of line 903, and possibly one or two other places in the pull request. I'd suggest the pull request probably needs to be rebased on those recent changes? Otherwise this patch re-introduces the bugs fixed when preferredLink strings are malformed in the source feed.

babbage (commented):
I am only a recent contributor to NetNewsWire and have not been given any sort of mandate to review other people's pull requests. So firstly this is not your official review by the NNW team, there will be other people who look at this. But a couple of quick comments.

Fixing a couple of other recent bugs that were triggered by malformed or unexpected feeds caused me to be looking at a number of site's raw RSS and JSON feeds directly. What I learned in this is that while the specs may be fairly clear on what url and externalURL are expected to mean in those feeds, we can't assume that they are always used this way. For instance, SixColors.com has Linked Item articles that are primary links to external sites, with some commentary, and other articles that are more primarily articles on their own site. You might think that those Linked Item articles would have the SixColors.com url in the "url" field and the external linked article in the "externalURL" field of the feed—but they do not. Instead, the external linked article is in the "url" field.[^1]

I think this has relevance to this patch. I suggest that we want to make as few assumptions as possible in the interface strings about what these fields link to. One way to do this would be that if there is only a single URL, whichever field it is in, the interface string should just be "Copy URL". Only when both the url and externalURL fields both have content should the text then be separated into "Copy Article URL" and "Copy External URL"—or whatever strings are selected.

I've also made one comment in the code as it relates to a recent patch, given this pull request is against a different branch at present.

[^1]: I think the SixColors feed is set up this way because in some cases their linked items are solely the first paragraph of an external article with "Continue reading on MacWorld.com..." so it would make no sense to send the user from an RSS reader to the SixColors.com short extract and then they have to click again to read the actual article. But they have other Linked Items that are not like this, and still it links only to the external article in their feed—I think due to limitations in how much they can alter what is in their feed setup.

iOS/MasterTimeline/MasterTimelineViewController.swift
babbage: Actually let me correct the above. It doesn't re-introduce the bug per se, it is just vulnerable to the issue which means occasionally it will fail to be able to share a link that otherwise would have been shared. All other places this has been fixed in the code base but this would become anew one not fixed.

vincode-io: We have since back merged the changes in `mac-release` and `ios-candidate` to `main`. This means that `Artilcle.preferredURL` is now on `main` and @babbage 's suggest change can be made.

robmathers: Thanks for the update. I'll rebase and adjust to use `preferredURL`

PR #3032, opened 2021-04-22
Adds —deep code sign flag for Xcode 12.5

stuartbreckenridge:
This allows the Mac target to build when using Xcode 12.5. The README and setup script have been updated. This will close #3031

Issue #3031, opened 2021-04-21
Code Sign Issues on Xcode 12.5


stuartbreckenridge:
The Mac target cannot build due to a Code Signing error related to the Share Extension:

```
/Users/stuart/Library/Developer/Xcode/DerivedData/NetNewsWire-gdvxpmwdeprglrcuiptwdqhzimnb/Build/Products/Debug/NetNewsWire Share Extension.appex/Contents/Frameworks/SyncDatabase.framework/Versions/A: code object is not signed at all
In subcomponent: /Users/stuart/Library/Developer/Xcode/DerivedData/NetNewsWire-gdvxpmwdeprglrcuiptwdqhzimnb/Build/Products/Debug/NetNewsWire Share Extension.appex/Contents/Frameworks/SyncDatabase.framework/Versions/A/Frameworks/RSCore.framework
Command CodeSign failed with a nonzero exit code
```

stuartbreckenridge:
Potential solution is to add the following to the DeveloperSettings file:

`OTHER_CODE_SIGN_FLAGS = --deep`

I am not sure of the side effects, however.

heckj:
FWIW:

`--deep` recursively signs code and libraries within a framework - it's actually specifically meant for this sort of scenario: where you have multiple layers in a framework that all need to be signed for execution validation. What changed in 12.4 -> 12.5 is a definitely curious scenario (worthy of a FB as a regression). But in general, I think adding in `--deep` is pretty safe for the project overall. Worst case scenario, you're signing more code that gets explicitly included into the built framework. (Only docs I could find for this from Apple were at <https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG404>) 

My suspicion is that the advice in that TN article to not use `--deep` as a general action is to prevent someone from slipping in code on a developer, and then you "accidentally" sign it, as a potential security hole.

PR #3129, opened 2021-05-23
Solve codesign & library embedding for Share Extension on Xcode 12.5

GetToSet:
A codesign issue appears on clean build with Xcode 12.5. #3031

Screenshot 2021-05-23 AM 9 21 44

Seems that RSCore was copied into wrong location, and adding a `--deep` (#3032) is not the ideal solution.

Screenshot 2021-05-23 AM 9 33 51

I found that RSCore should be an explicit dependency of SyncDatabase, which is [imported](https://github.com/Ranchero-Software/NetNewsWire/blob/7558120dc39004e9d57e02c3e36c20fb4223358b/SyncDatabase/Sources/SyncDatabase/SyncDatabase.swift#L10) in related code.

This pull request adds RSCore as an explicit dependency of SyncDatabase to solve codesign issue on Xcode 12.5.


stuartbreckenridge:
The only problem I see with this solution is that the Share Extension is ~10x the size it was before:

Screenshot 2021-05-23 at 21 46 27

GetToSet:
> The only problem I see with this solution—though I doubt adding this dependency is the culprit—is that the Share Extension is ~10x the size it was before. (This appears to be an Xcode 12.5 problem in the way it is determining where frameworks should go.)
> 
> Screenshot 2021-05-23 at 21 46 27

I'll install a previous version of Xcode and take a deeper investigation to see what's actually changed.

stuartbreckenridge:
There are two regressions I've identified in Xcode 12.5:
- The first, as you've got in your [screenshot](https://user-images.githubusercontent.com/8158163/119262692-acd2eb00-bc0e-11eb-8857-49f728bfc4e5.png), is that the Sparkle framework is copied _after_ the Code Sign Sparkle Run Script. As a result, the script can't remove unneeded components because they are not in the location expected. In Xcode 12.4 and earlier, that ordering is reversed, i.e. Copy then Script, and it allows the Run Script to remove unneeded components after to the Copy.
- Xcode 12.5, with our current Package configuration, embeds all the Frameworks in the Share Extension. (Hence the ballooning in size!)

GetToSet:
For Xcode 12.4: 
* Dynamic library decencies of 'Account', which NetNewsWire Share Extension relies on, won't be copied into bundle when it's not listed in 'Embed Frameworks'.
* Linking against 'RSCore' automatically adds it into 'Embed Frameworks', and removing it also removes it from linking phase. It seems a Xcode bug, 'RSCore' has to be bundled though it's not needed and therefore a script was needed to remove RSCore from the final bundle.

https://user-images.githubusercontent.com/8158163/119300228-febd5480-bc92-11eb-9b68-ea857243df23.mov

For Xcode 12.5:
* Removing 'RSCore' in 'Embed Frameworks' won't remove it from being linked, therefore the deletion script is not needed any more. (The first commit, it's backwards compatible with Xcode 12.4)
* Swift Package Manager 5.4 in Xcode 12.5 seems to have new behavior, which forces copy and sign dynamic frameworks that 'Account' relies on even they're not included in 'Embed Frameworks', in a very late process, even after cleanup script runs. Possibly related to [this](https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-release-notes#Swift-Packages).

Screenshot 2021-05-24 AM 10 55 40

* Moving cleanup script into the main app build phase should be the simplest way to solve this problem, which is what the second commit did.

GetToSet:
For issues related with Sparkle on Xcode 12.5, maybe i'll open another PR for this.

vincode-io:
@GetToSet What's the status of this pull request given that we are currently staying on Xcode 12.4 for development. Does it have any net benefit?

GetToSet:
> @GetToSet What's the status of this pull request given that we are currently staying on Xcode 12.4 for development. Does it have any net benefit?

Initially I found the project won't build on Xcode 12.5 and try to fix it. There's no need for this PR if only Xcode 12.4 is targeted.

vincode-io:
@GetToSet I've tested this pull request and want to merge it. Could you please resolve the merge conflicts so that I can?

GetToSet:
> @GetToSet I've tested this pull request and want to merge it. Could you please resolve the merge conflicts so that I can?

Seems this PR has already been merged.

Issue #3123, opened 2021-05-19
Tweetbot share sheet is empty 


stuartbreckenridge:
When invoking the Tweetbot share sheet on the Mac or iOS, there is no content. 

- [ ] macOS
- [x] iOS (fixed via #3128) 

https://user-images.githubusercontent.com/7046652/118808998-aaa22100-b8dc-11eb-8999-5e2eda1daad0.mov

PR #3130, opened 2021-05-24
Improves compatibility with sharing services

stuartbreckenridge:
In relation to #3123: the changes in this commit allow text to be shared to additional apps in the share menu.

The first change is the ordering of the `writableTypes` array. It now puts `.string` first. This change on its own allows text to be shared to Tweetbot, Ulysees, and potentially other apps.

The second change is to the text—`plainText()`—that is shared. It has been reduced to `title` and `url`, which makes more sense in the context of sharing snippets. This is in large part to do with the fact that sharing title, contentText, summary, contentHTML, url, externalURL, date, and feedName is somewhat overkill when sharing to a Twitter-like service.

Lastly, the downside to this change is that Add to Reading List appears to have disappeared.


vincode-io:
@stuartbreckenridge What's the status of this pull request?

vincode-io:
I'm closing this draft proposal since it is so old. Let's submit it again if we want to move forward with it.

PR #3152, opened 2021-06-13
Multiplatform | SwiftUI 3

stuartbreckenridge:
• Targets bumped to iOS 15 and macOS 12.
• Adds `IconImageCache` to the Multiplatform targets and fixes unwrap. 
• New `.syntax` used for styles instead of `Style()`
• Fixes vertical stutter on article view during window resize.
• `@AppStorage` properties in `AppDefaults` have had `objectWillChange` removed.
• Custom `MacSearchField` file deleted
• Adds placeholder `.searchable` search field.
• Toolbar navigation subtitle now has unread count.
• Toolbar navigation title now matches AppKit behaviour.
• Removes (lack of) ability to set default RSS feeder option from preferences
• Fixes a crash in Account preferences view
• Toolbars reconfigured to match AppKit
• Removes filter from top of timeline to match AppKit
• `.onAppear{}` changed to `.task{}`
• `.withAnimation` used to show the refresher.


vincode-io:
DEPLOYMENT_TARGET changes should be made in the xcconfig files. Maybe we don't have the VerifyBuildSettings script running on the Multiplatform targets? It should have warned you about this.

Could you change this to merge to the new `swiftui3` branch I just created? That way we can validate that SwiftUI 3 is going to cut it for us this year.

stuartbreckenridge:
@vincode-io I've updated the base branch and the `.xcconfig` files. However, if I reset the Mutliplatform targets to a state where they don't trigger VerifyNoBuildSettings, then there is a whole set of other warnings about iOS 14/15 features not being available.

PR #3154, opened 2021-06-14
Multiplatform/swiftui3

stuartbreckenridge:
• Targets bumped to iOS 15 and macOS 12.
• Adds `IconImageCache` to the Multiplatform targets and fixes unwrap. 
• New `.syntax` used for styles instead of `Style()`
• Fixes vertical stutter on article view during window resize.
• `@AppStorage` properties in `AppDefaults` have had `objectWillChange` removed.
• Custom `MacSearchField` file deleted
• Adds placeholder `.searchable` search field.
• Toolbar navigation subtitle now has unread count.
• Toolbar navigation title now matches AppKit behaviour.
• Removes (lack of) ability to set default RSS feeder option from preferences
• Fixes a crash in Account preferences view
• Toolbars reconfigured to match AppKit
• Removes filter from top of timeline to match AppKit
• `.onAppear{}` changed to `.task{}`
• `.withAnimation` used to show the refresher.


vincode-io:
It looks like there are still build settings changes for version numbers in the `project.pbxproj` file.

stuartbreckenridge:
Without manually setting the targets to iOS 15 and macOS 12, they won't build (even with the target versions set in `.xcconfig` to 15/12). You end up with errors:

> 'banner' is only available in macOS 11.0 or newer
> init(wrappedValue:_:store:)' is only available in macOS 11.0 or newer

stuartbreckenridge:
I think it's the target version override in the multiplatform `xcconfig` that isn't being respected.

vincode-io:
Have you tried manually deleting the buildSettings from `project.pbxproj` using a text editor?

stuartbreckenridge:
That seems to fix it, thanks!

PR #3155, opened 2021-06-15
iOS now uses `refreshable`

stuartbreckenridge:
AccountManager now has an async function for refreshing accounts (for review)

### DIFF
--- a/Account/Sources/Account/AccountManager.swift
+++ b/Account/Sources/Account/AccountManager.swift
@@ -296,40 +296,11 @@ public final class AccountManager: UnreadCountProvider {
 		
 	}
     
-    @available(iOS 15, *)
-    @available(macOS 12, *)
     /// Asyncronous refresh of all accounts.
-    /// - Parameter withEarlyContinuation: if `true`, the function will return immediately, if `false`, the function will return when the refresh is complete.
-    public func refreshAll(withEarlyContinuation: Bool = true) async throws {
-        return try await withUnsafeThrowingContinuation { continuation in
-            let group = DispatchGroup()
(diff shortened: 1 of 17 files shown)


vincode-io:
I'm really behind on the WWDC sessions for concurrency, so correct me if I'm wrong. Shouldn't we be throwing something now instead of returning the tuple `AsyncRefreshResult`?

stuartbreckenridge:
I thought about using `withUnsafeThrowingContinuation` but the problem is that a continuation can only run once. 

If you have three accounts and the second of those has an error, you can't throw an error there and continue on to refresh the third, because you've used the continuation already.

stuartbreckenridge:
fb500ce: `refreshAll()` will now throw in situations where there is an connection error. Other refresh errors are not surfaced.

In refreshable it's now `try? await AccountManager.shared.refreshAll()`

brentsimmons:
I recall that NetNewsWire was one of the apps that people were using to work on concurrency — is that fork somewhere around, still? Is it worth looking at? (Maybe we already are?)

stuartbreckenridge:
https://github.com/kavon/NetNewsWire/pull/1

stuartbreckenridge:
I will take a swing at doing this without using continuations.

Account/Sources/Account/AccountManager.swift
```diff
@@ -12,6 +12,9 @@ import RSWeb
 import Articles
 import ArticlesDatabase
 
+public typealias AsyncRefreshResult = (Bool, [Error]?)
+
+
```
vincode-io, line 17: This can go away now, right?

PR #3160, opened 2021-06-16
[iOS] Add/Update Account Can Use 1Password 

stuartbreckenridge:
• Adds 1Password provided `.h` + `.m` files and creates Bridging Header 
• Adds 1Password nav bar image asset, tints it using .accentColor
• Fixes Activity Spinner positioning on the nav bar
• 1Password **will only be visible** if the 1Password app is available


vincode-io:
It looks like some build settings got introduced into the `project.pbxproj` file.

stuartbreckenridge:
That should be it fixed.

PR #3262, opened 2021-08-23
Adds option to open article links in NetNewsWire or in the default browser.

### LINKED ISSUE #3029: iOS Add option to open all links in default browser
Wonderful app! I'm wondering if you would consider adding an option to have links open in the default system browser instead of the in-app browser?

stuartbreckenridge:
- Adds `BrowserManager`, `Browser`, `BrowserConfigurationViewController`, and appropriate browser glyphs.
- The default behaviour will allow a user to to select NetNewsWire (in-app, default) or Default Browser.
- If we want to override the default browser, it can be enabled by setting `BrowserManager`'s `enableExtendedBrowserPreferences` to `true`. This then opens up options for Brave, Chrome, Edge, Firefox, Opera and 1Password.
- The Article section in Settings has been rearranged so that on iPad the second row displays the browser preference instead of the `Enable Full Articles` option.

Fixes #3029


brentsimmons:
I’d like for this to be simplified to only what we need to give users a choice between in-app and default browser. Let’s not include extra code for specifying some other browser — we may not ever want to do that feature.

stuartbreckenridge:
@brentsimmons fixed in 3bc18e2.

PR #3301, opened 2021-09-23
Theme tweaks

stuartbreckenridge:
On macOS:
- The alert wording has been changed to "This theme cannot be used because _x_ is missing from the Info.plist".
- The alert contains an additional button that opens up the folder of the theme that the user was attempting to open or import. (When clicked, this button does not dismiss the alert.)

On iOS:
- The alert wording has been changed to "This theme cannot be used because _x_ is missing from the Info.plist".

### DIFF
--- a/Mac/AppDelegate.swift
+++ b/Mac/AppDelegate.swift
@@ -926,9 +926,9 @@ internal extension AppDelegate {
 			case .typeMismatch(let type, _):
 				informativeText = "the type—'\(type)'—is mismatched."
 			case .valueNotFound(let value, _):
-				informativeText = "the value—'\(value)'—is not found."
+				informativeText = "the value—'\(value)'—is missing in the Info.plist."
 			case .keyNotFound(let codingKey, _):
-				informativeText = "the key—'\(codingKey.stringValue)'—is not found."
+				informativeText = "the key—'\(codingKey.stringValue)'—is missing in the Info.plist."
 			case .dataCorrupted( _):
 				informativeText = error.localizedDescription
 			default:
--- a/Mac/AppDelegate.swift
+++ b/Mac/AppDelegate.swift
@@ -107,6 +107,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSUserInterfaceValidations, 
 	private var softwareUpdater: SPUUpdater!
 	private var crashReporter: PLCrashReporter!
 	#endif
+	
+	private var themeImportPath: String?
 
 	override init() {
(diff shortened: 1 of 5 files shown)


Mac/AppDelegate.swift
```diff
@@ -921,16 +919,16 @@ internal extension AppDelegate {
...
+				informativeText = "the type—'\(type)'—is mismatched."
 			case .valueNotFound(let value, _):
-				informativeText = "Value '\(value)' not found."
+				informativeText = "the value—'\(value)'—is missing in the Info.plist."
 			case .keyNotFound(let codingKey, _):
-				informativeText = "Key '\(codingKey.stringValue)' not found."
+				informativeText = "the key—'\(codingKey.stringValue)'—is missing in the Info.plist."
 			case .dataCorrupted( _):
 				informativeText = error.localizedDescription
 			default:
```
vincode-io, line 934: We should use `NSString.localizedStringWithFormat` here instead of string interpolation in case we do end up localizing the AppKit version of the app. I would normally not worry about it, but this would be easily overlooked if we did do localization on the AppKit version.

iOS/UIKit Extensions/UIViewController-Extensions.swift
```diff
@@ -19,13 +19,13 @@ extension UIViewController {
...
 			case .valueNotFound(let value, _):
-				let str = "Value '\(value)' not found."
+				let str = "This theme cannot be used because the value—'\(value)'—is missing in the Info.plist."
 				presentError(title: errorTitle, message: str, dismiss: dismiss)
 			case .keyNotFound(let codingKey, _):
-				let str = "Key '\(codingKey.stringValue)' not found."
+				let str = "This theme cannot be used because the key—'\(codingKey.stringValue)'—is missing in the Info.plist."
 				presentError(title: errorTitle, message: str, dismiss: dismiss)
 			case .dataCorrupted( _):
 				presentError(title: errorTitle, message: error.localizedDescription, dismiss: dismiss)
```
vincode-io, line 31: We should use `NSString.localizedStringWithFormat` here.

Mac/AppDelegate.swift
stuartbreckenridge: Sorted.

iOS/UIKit Extensions/UIViewController-Extensions.swift
stuartbreckenridge: Sorted.

PR #3307, opened 2021-09-30
Use repaired URLs for link, externalLink and imageLink where needed. Fixes #3103

### LINKED ISSUE #3103: Copy URLs using repaired rather than raw feed links
Looking at #3102 today made me reflect that we have recently added the capacity to repair article URLs that are malformed through containing strings that are not percent encoded, and we are considering whether to also extend this to removing `CDATA` tags from these fields too in how to fix #3043. 

We are only fixing link strings where they cannot be converted into a valid URL. If we use the raw feed url string when the user is copying or sharing URLs, we are handing to the user a link we know is invalid. In some cases the destination app will resolve the issue—such as when pasting into Safari a link that contains spaces that have not been percent encoded. But in other cases the user ends up with something that does not work, even though the outgoing link works in NetNewsWire.

I suggest a patch that converts the various copy URL and share facilities to use the string-correcting pipeline that was established in #3079. This will involve adding a URL.init guard step into some places where previously the raw feed string was just being passed without that, so we would want to be mindful of performance and only use it where required.

Would make sense to wait for #3102 to land then I'd be happy to take a look at this.

babbage:
As discussed in #3103 and in Slack, this approach provides computed properties on `Article` of `link`, `externalLink`, `imageLink` based on associated URL properties of `url`, `externalURL` and `imageURL`. The properties that expose the underlying stored string values are renamed to `rawLink`, `rawExternalLink`, and `rawImageLink` _on Article_. No change is made to the database API nor database storage, and the init method for Article continues to use `url:`, `externalURL:` and `imageURL:` which are provided with String? values. 

These changes ensure that where stored link, externalLink and imageLink properties require repair and can be repaired, these repaired links will be used. However, URLs are only created, and if necessary repaired, when these computed variables are called, so the change has no performance impact on Article loading. The computed properties for link fall back to providing the raw value stored from the original feed in the event a URL cannot be parsed from the link.

At present the only repair attempted on links is to replace unescaped string characters with percent encoding. However, this has been abstracted into `URL.reparingIfRequired(_ String?)` so there is one point for future changes should this be identified as desirable by future errors in feed links that NNW wishes to automatically repair. 

The first commit implements the renaming of the existing Article properties, and the provision of the new computed properties, while maintaining the direct use of the raw stored values as previously. The second commit implements the use of the repaired link and URL properties.

### DIFF
--- a/Account/Sources/Account/CloudKit/CloudKitArticlesZone.swift
+++ b/Account/Sources/Account/CloudKit/CloudKitArticlesZone.swift
@@ -223,10 +223,10 @@ private extension CloudKitArticlesZone {
 		record[CloudKitArticle.Fields.title] = article.title
 		record[CloudKitArticle.Fields.contentHTML] = article.contentHTML
 		record[CloudKitArticle.Fields.contentText] = article.contentText
-		record[CloudKitArticle.Fields.url] = article.url
-		record[CloudKitArticle.Fields.externalURL] = article.externalURL
+		record[CloudKitArticle.Fields.url] = article.rawLink
+		record[CloudKitArticle.Fields.externalURL] = article.rawExternalLink
 		record[CloudKitArticle.Fields.summary] = article.summary
-		record[CloudKitArticle.Fields.imageURL] = article.imageURL
+		record[CloudKitArticle.Fields.imageURL] = article.rawImageLink
 		record[CloudKitArticle.Fields.datePublished] = article.datePublished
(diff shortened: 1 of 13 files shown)


babbage:
Oh, a postscript here from my point of view is that this also slightly cleans up the API for engaging with a property, as now all `String?` properties that refer to a location on the internet are some form of `Article.*link` while all properites that are in the form of `Article.*URL` actually return `URL` types.

vincode-io:
I like you you made the API more consistent. Thanks @babbage!

Issue #3280, opened 2021-09-08
Content style: `<a>` element containing `<code>` doesn't look like a hyperlink


danburzo:
<img width="642" alt="Screenshot 2021-09-08 at 10 42 06" src="https://user-images.githubusercontent.com/205375/132467476-4c47b7d4-cab1-4ec8-9385-73dab42ba762.png">

The article in question is: https://danburzo.ro/javascript-codemods/

danburzo:
Awesome! Thank you.

PR #3312, opened 2021-10-05
Make sure <code> inside <a> uses the link color

Wevah:
It's not obvious when a link containing only a `<code>` element is in fact a link with our current default style sheet.

Fix for #3280.


Wevah:
I couldn't think of any textual elements that we currently style that we'd want to show up without link color inside links, but I'd welcome feedback.

PR #3339, opened 2021-11-06
Fixes #3335

### LINKED ISSUE #3335: iOS: Specific views should dismiss when the app is restored via the widget or a notification
If the user opens the app via a tap on an article in the widget or through a notification and the Settings view or an SFSafariViewController is the topmost view, those views do not dismiss to make way for the selected content. Ideally, those views should dismiss, otherwise, it looks like the app isn’t doing what the user asked.

stuartbreckenridge:
When the app is brought to the foreground from an external action (e.g., tapping on the widget, opening from a notification), a notification is posted (with a slight delay).

`MasterFeedViewController` and `SettingsViewController` are observers. `MasterFeedViewController` will dismiss any `SFSafariViewController`s that are presented, while `SettingsViewController` will dismiss itself. With these tweaks, the existing behaviour in #3335—where existing views will obscure what the user has requested—is fixed.

### DIFF
--- a/iOS/AppDelegate.swift
+++ b/iOS/AppDelegate.swift
@@ -211,9 +211,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 			if let sceneDelegate = response.targetScene?.delegate as? SceneDelegate {
 				sceneDelegate.handle(response)
 				DispatchQueue.main.asyncAfter(deadline: .now() + 0.5, execute: {
-					NotificationCenter.default.post(name: .DidLaunchFromExternalAction, object: nil)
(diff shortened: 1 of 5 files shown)


vincode-io:
Do we have to add a notification to do this? I think we could just have the `SceneCoordinator` tell `MasterFeedViewController` to dismiss anything it is showing.

`SceneDelegate` -> `SceneCoordinator` -> `MasterFeedViewController`

stuartbreckenridge:
I will look at it again tomorrow.

stuartbreckenridge:
@vincode-io That last commit changes the design to use a function on SceneCoordinator. The delay is retained because animating immediately is objectively not nice to look at. 🙃

Issue #3233, opened 2021-07-25
Feature Request: Feed Activity Dashboard


MDWolinski:
One feature I miss from using Feedbin is the ability to see a dashboard of feeds with an activity line chart of when articles were last received.  This was a great way to see how active (or not) a feed was and to be able to clean-up feeds that haven't been updated in awhile.

Think this could be a great feature to have

brentsimmons:
NetNewsWire of Yore has had features for helping to find dinosaurs.

vincode-io:
Do we want to mark this as a duplicate of #1107?

MDWolinski:
I’m cool with this being a duplicate. Didn’t see it in my search but obviously different view on what to call it.

vincode-io:
Duplicate of #1107.

PR #3340, opened 2021-11-08
Execute WidgetDataEncoder synchronously before suspending database

dkadams:
Refactor WidgetDataEncoder to provide both sync & async entry points.
Update callers that subsequently suspend the database to encode widget
data synchronously. This prevents the encoding from blocking on the
suspended database.

Partial fix for Ranchero-Software#3200

### DIFF
--- a/Shared/Widget/WidgetDataEncoder.swift
+++ b/Shared/Widget/WidgetDataEncoder.swift
@@ -29,73 +29,79 @@ public final class WidgetDataEncoder {
 	private init () {}
 	
 	@available(iOS 14, *)
-	func encodeWidgetData() throws {
+	func encodeWidgetDataSync() throws {
 		
 		os_log(.debug, log: log, "Starting encoding widget data.")
-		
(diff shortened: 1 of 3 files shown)


vincode-io:
I think all we need to do is back out #3323 don't you? I don't understand what an async version buys us.

dkadams:
I agree that backing out #3233 should prevent widget data encoding from blocking due to the suspended db. But that PR also removed some lines that called `UIApplication.shared.endBackgroundTask` and I'm not really sure what reintroducing that logic will do.

I only added the async version to support the `SceneDelegate` call site, which didn't appear to be suspending the database. I'm not sure what triggers `sceneDidEnterBackground` and was worried that calling the widget data encoding synchronously might be problematic.

vincode-io:
I decided that it would be cleaner to just back out  #3323, so we don't need to merge this pull request.

@dkadams , your help on this is greatly appreciated. We understand the root cause a lot better now. I'm not sure what to do about it yet...

PR #3359, opened 2021-11-24
Implement pull up mark as read

everhardt:
WIP: only iOS target is working.

* Added the pull up as UITableViewController implementation
* Improved marking as read on scroll: register top visible article when scrolling starts and ignore the ones above that

### DIFF
--- a/Mac/Base.lproj/Preferences.storyboard
+++ b/Mac/Base.lproj/Preferences.storyboard
@@ -31,23 +31,31 @@
         <scene sceneID="R4l-Wg-k7x">
             <objects>
                 <viewController title="General" storyboardIdentifier="General" id="iuH-lz-18x" customClass="GeneralPreferencesViewController" customModule="NetNewsWire" customModuleProvider="target" sceneMemberID="viewController">
-                    <view key="view" misplaced="YES" id="WnV-px-wCT">
-                        <rect key="frame" x="0.0" y="0.0" width="509" height="338"/>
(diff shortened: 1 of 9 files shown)


everhardt:
It should also probably be disabled when `!AppDefaults.shared.markArticlesAsReadOnScroll`

everhardt:
Current state:
* fully working for iOS: mark as read on scroll _and_ pull up from bottom of the article list to mark the last few articles as read
* partially working for macOS: only mark as read on scroll

I failed to build a good control for macOS to pull up from the bottom of the article list to mark the last few articles as read. I spent quite some time figuring out how to get a view behind the NSTableView as that does not have a backgroundView property like its UITableView sibling. Anyway, I found a workaround. The problem I then faced is that the elasticity of the NSScrollView has too much resistance for what we want: I'm often not able to scroll it up far enough for the text "Pull up to mark as read" to show, let alone far enough for it to trigger marking it as read. AFAIK there's no way to decrease that resistance, other than to rebuilt the whole scrolling thing (a bit like https://github.com/alexzielenski/ScrollToRefresh). I propose to go for a solution that fits the macOS UX better and the best I can come up with is a row at the end of the articles list with a button "Mark all as read". Granted, it is pretty redundant given that we have such a button at the toolbar already, but if you read from top to bottom it fits the flow of the user better.

vincode-io:
We've decided to not move forward with this feature. See the discussion in the #work Slack work channel or ask @brentsimmons if you have any questions.

PR #3415, opened 2022-01-23
iOS UI Refresh

stuartbreckenridge:
- [x] Feeds screen

### DIFF
--- a/iOS/MasterFeed/Cell/MasterFeedTableViewSectionHeader.swift
+++ b/iOS/MasterFeed/Cell/MasterFeedTableViewSectionHeader.swift
@@ -85,7 +85,7 @@ class MasterFeedTableViewSectionHeader: UITableViewHeaderFooterView {
 
 	private lazy var disclosureButton: UIButton = {
 		let button = NonIntrinsicButton()
-		button.tintColor = AppAssets.primaryAccentColor
+		button.tintColor = AppAssets.secondaryAccentColor
 		button.setImage(AppAssets.disclosureImage(size: 14, weight: .bold), for: .normal)
 		button.contentMode = .center
 		button.addInteraction(UIPointerInteraction())
--- a/iOS/MasterFeed/Cell/MasterFeedTableViewSectionHeader.swift
+++ b/iOS/MasterFeed/Cell/MasterFeedTableViewSectionHeader.swift
@@ -77,7 +77,7 @@ class MasterFeedTableViewSectionHeader: UITableViewHeaderFooterView {
 		label.numberOfLines = 0
 		label.allowsDefaultTighteningForTruncation = false
 		label.adjustsFontForContentSizeCategory = true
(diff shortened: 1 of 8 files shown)


stuartbreckenridge:
### Feeds

- The Feeds screen now looks more like News, with bold section headers and the disclosure icon moved to the left.
- Folders are bold, with nested items indented to convey the hierarchy. 

nnw-ui

gferreroferri:
Should't folders show the folder icon like the macOS version to keep consistency instead of the > indicator?

stuartbreckenridge:
On the Mac, we show chevron + folder icon + feed. We may go that route here, but my thinking here is that we are space-constrained with fixed sidebar sizes and showing three may be overkill.

gferreroferri:
probably the chevron would not be needed if the folder icon changes between close folder and open folder icon

vincode-io:
@stuartbreckenridge is correct. We don't show the folder icon because of space constraints.

iOS/MasterFeed/Cell/MasterFeedTableViewSectionHeader.swift
```diff
@@ -139,7 +139,6 @@ private extension MasterFeedTableViewSectionHeader {
 	}
 	
 	func commonInit() {
-		addSubviewAtInit(unreadCountView)
```
vincode-io, line 142: I'm assuming the rest of the clean up happens in another commit?

stuartbreckenridge: e4cb868

vincode-io: Thanks!

iOS/MasterTimeline/MasterTimelineViewController.swift
```diff
@@ -1052,7 +1052,7 @@ private extension MasterTimelineViewController {
 		let webFeed = coordinator.timelineFeed as! WebFeed
 		let homePageUrlTitle = NSLocalizedString("Copy Home Page URL", comment: "Copy Home Page URL")
 		let homePageCopyURLAction = UIAction(title: homePageUrlTitle,
-											 image: AppAssets.globeImage,
+											 image: AppAssets.safariImage,
```
vincode-io, line 1055: Shouldn't these icons match their direct equivalents on the Sidebar context menu?

stuartbreckenridge:
The PR only includes sidebar changes

iOS/MasterTimeline/MasterTimelineViewController.swift
stuartbreckenridge: The do now, but I've removed the context menu work from this PR.

gferreroferri:
Could it be done following Mail.app guidelines? 
0D6F65E7-C783-401B-B3B3-A1CF634CDB65

PR #3441, opened 2022-02-04
Notifications Manager Performance Improvements

stuartbreckenridge:
Adopts prefetch for smooth scrolling, instead of relying on notifications. 

(Tested with iOS Dev Directory.)


vincode-io:
Why don't you have code to respond to the image available notifications? If the image wasn't in the `IconImageCache` then you only get the default globe image for the feed and it is never updated.

PR #3462, opened 2022-02-12
Adds unread count indicator to Article view

### LINKED ISSUE #3134: [Feature request] Unread count on article view
Could an unread count be added to the header in article view? As one swipes through unread articles, this count would decrease.   

It would help me to understand how far through my unread articles I am.

I've attached a quick mockup.

Many thanks for considering this!

image

stuartbreckenridge:
Fixes #3134

### DIFF
--- a/iOS/Article/ArticleViewController.swift
+++ b/iOS/Article/ArticleViewController.swift
@@ -105,6 +105,8 @@ class ArticleViewController: UIViewController, MainControllerIdentifiable {
 			parentNavController.interactivePopGestureRecognizer?.delegate = poppableDelegate
 		}
 		
+		navigationItem.leftItemsSupplementBackButton = true
+		
 		pageViewController = UIPageViewController(transitionStyle: .scroll, navigationOrientation: .horizontal, options: [:])
 		pageViewController.delegate = self
 		pageViewController.dataSource = self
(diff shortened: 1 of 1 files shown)


vincode-io:
If someone is reading an article while a sync is happening, the unread count in the timeline could change.

You might want to add an API function to the ArticleViewController that the SceneCoordinator can call when an Article fetch completes. That would be more efficient than just handing the unread count did change notification.

stuartbreckenridge:
@vincode-io Is that last commit more what you had in mind?

vincode-io:
Yes, exactly!

stuartbreckenridge:
I have gated in the unread count update so it only appears on the iPhone. It seems overkill on iPad as the timeline is visible.

vincode-io:
Good call.

londondev77:
I originally requested this feature - many thanks for implementing! Can I be a disenting voice and say that on iPad it would still be useful?  Maybe there can be a preference to show/hide the count?

stuartbreckenridge:
It's implemented on iPad when in portrait (added in a later commit). 

When in landscape, the timeline view is there and the count there.

londondev77:
Brilliant - thanks!

londondev77:
Looks like this has been removed. Any chance of reintegrating?

PR #3517, opened 2022-04-04
Add proper OSLog argument for os_log calls.

GetToSet:
The [`os_log`](https://developer.apple.com/documentation/os/3580306-os_log) function defaults `log` parameter to [`OSLog.default`](https://developer.apple.com/documentation/os/oslog/2320724-default) which has no `category` and `subsystem` associated. This makes it extremely difficult to distinguish NetNewsWire logs from system ones within Console.app.

This PR adds proper OSLog argument for all os_log calls.

### DIFF
--- a/Mac/MainWindow/Timeline/TimelineViewController.swift
+++ b/Mac/MainWindow/Timeline/TimelineViewController.swift
@@ -940,7 +940,7 @@ extension TimelineViewController: NSTableViewDelegate {
 				return [action]
 
 			@unknown default:
-				os_log(.error, "Unknown table row edge: %ld", edge.rawValue)
+				print("Unknown table row edge: \(edge.rawValue)")
 		}
 
 		return []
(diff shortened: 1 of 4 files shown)


Mac/MainWindow/Timeline/TimelineViewController.swift
```diff
@@ -940,7 +940,7 @@ extension TimelineViewController: NSTableViewDelegate {
 				return [action]
 
 			@unknown default:
-				os_log(.error, "Unknown table row edge: %ld", edge.rawValue)
+				print("Unknown table row edge: \(edge.rawValue)")
```
brentsimmons, line 943: Is there a reason this is a print instead of os_log? Could this be os_log too?

brentsimmons (changes requested):

Mac/MainWindow/Timeline/TimelineViewController.swift
GetToSet: It seems most viewControllers use `print` instead of `os_log` for logging. I'm wondering if there's any reason to opt out `os_log` for UI-related codes and should I replace all those `print`s to `os_log`s or keep them as-is?

brentsimmons: Good point. Since we already have the precedent of using print for view controllers, let’s keep using print. Which means this change is the right thing to do, and I no longer am requesting changes. :)

PR #3519, opened 2022-04-05
Fix AccountTests build issue.

GetToSet:


Articles/Sources/Articles/Article.swift
```diff
@@ -20,7 +20,7 @@ public struct Article: Hashable {
 	public let contentHTML: String?
 	public let contentText: String?
 	public let rawLink: String? // We store raw source value, but use computed url or link other than where raw value required.
-    public let rawExternalLink: String? // We store raw source value, but use computed externalURL or externalLink other than where raw value required.
+	public let rawExternalLink: String? // We store raw source value, but use computed externalURL or externalLink other than where raw value required.
```
vincode-io, line 23: Did you change spaces to a tab or tab to spaces here? Regardless, let's revert this so that we don't have just a whitespace change in Article.swift please.

GetToSet: Sorry that I didn't notice this change. I'll revert this.

GetToSet:
I've updated this PR and rebased onto the latest `main` branch commit.

vincode-io:
The la for the fix!

PR #3584, opened 2022-06-03
fix #3580 - dragging between windows doesn't work

### LINKED ISSUE #3580: macOS - drag a feed from iCloud list to a folder in iCloud list in another NNW window fails
Context: Syncing with iCloud, macOS NNW Version 6.1 (6106), macOS 12.3.1 (21E258), 13" M1 MBP.

1. Open two NNW windows
2. drag a feed item from the top level of the iCloud group to a folder in the iCloud group in the second window.
-> Drag seems like it will work but the feed item just springs back.

haikusw:
Addresses #3580, see that issue for discovery and details but, basically, a view local property stored the nodes being dragged which was used in drag handling and that local property variable was nil in the copy of the view in another window so dragging between windows didn't work.

**NOTE:**  I think `draggedNodes` is an optimization that is flawed in a multi-window environment and I would like to remove it entirely.   This would reduce code complexity and overall amount of code.

I'd be happy to revise this PR with the change but it seemed a more radical change so I didn't want to do it without some feedback from the project maintainers.

### DIFF
--- a/Mac/MainWindow/Sidebar/SidebarOutlineDataSource.swift
+++ b/Mac/MainWindow/Sidebar/SidebarOutlineDataSource.swift
@@ -203,13 +203,23 @@ private extension SidebarOutlineDataSource {
 			return SidebarOutlineDataSource.dragOperationNone
 		}
 		if parentNode == dropTargetNode && index == NSOutlineViewDropOnItemIndex {
-			return localDragOperation(parentNode: parentNode)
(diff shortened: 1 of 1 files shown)


haikusw:
Ok, hold off.  Just saw that the //TODO: 's are still in here.

haikusw:
**This has introduced a somewhat subtle bug** , or possibly revealed it since cross windows drags are now available.  _Investigating_

PR #3586, opened 2022-06-04
Alternative to PR #3584 to fix #3580 

### LINKED ISSUE #3580: macOS - drag a feed from iCloud list to a folder in iCloud list in another NNW window fails

haikusw:
Alternative to PR #3584 to fix #3580.  This removes the `draggedNodes` cache optimization and always uses the drag pasteboard items for both within single window and across multiple window drags to reduce and simplify the code.

### DIFF
--- a/Mac/MainWindow/Sidebar/SidebarOutlineDataSource.swift
+++ b/Mac/MainWindow/Sidebar/SidebarOutlineDataSource.swift
@@ -203,13 +203,23 @@ private extension SidebarOutlineDataSource {
 			return SidebarOutlineDataSource.dragOperationNone
 		}
 		if parentNode == dropTargetNode && index == NSOutlineViewDropOnItemIndex {
-			return localDragOperation(parentNode: parentNode)
+			if draggedNodes?.isEmpty == false {
+				return localDragOperation(parentNode: parentNode)
+			}
+			else {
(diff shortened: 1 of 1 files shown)


haikusw:
I think this is a better approach than #3584 but they are equally ready for review.

There is one comment in there (twice) that you'll probably want removed.  I removed a bunch of logic that was unnecessary and it felt like an explanation was needed.  Not sure if you want to leave that in the code (doesn't seem to be the convention) or if there's an appropriate tech note to move that to, or maybe I should have added that as a comment on the diff of this PR as an aid to the reviewer?  Happy to learn how you prefer things to be done.

This could use testing by other folks, especially if you have a way to use other than just two local accounts as that's all I've been able to test cross account cross window drags.

haikusw:
Found a bug that reveals a deeper level of work needed.  Have that in-progress and will re-open this or make a new clean PR with it once I've got that working.

PR #3587, opened 2022-06-05
Address #3580 by avoiding use of per-window properties.

haikusw:
Address #3580 by using only the items in the DraggingInfo to handle the drag and no view properties associated with a single window.

Remove cached `draggedNodes` optimization and just always use the data on the dragging pasteboard because it works both within the same window and across two windows.

Add support for WebFeedPasteboardWriter to include a parent containerName in the pasteboard dictionary written for a WebFeed object (which doesn't know about it's parent) because we need this for drag and drop to be able to perform a move operation.

Fix issue with dragged feeds not knowing their parent folder and delete deleting the wrong item by telling WebFeedPasteboardWriter what the container of items is.  Remove code that is now unused now that draggedNodes is removed and consolidate code that was expanded to handle both node and pasteboard items since we are just using pasteboard items now.

### DIFF
--- a/Mac/MainWindow/Sidebar/PasteboardWebFeed.swift
+++ b/Mac/MainWindow/Sidebar/PasteboardWebFeed.swift
@@ -15,7 +15,7 @@ typealias PasteboardWebFeedDictionary = [String: String]
 
 struct PasteboardWebFeed: Hashable {
 
-	private struct Key {
+	fileprivate struct Key {
 		static let url = "URL"
 		static let homePageURL = "homePageURL"
 		static let name = "name"
@@ -25,6 +25,7 @@ struct PasteboardWebFeed: Hashable {
 		static let accountType = "accountType"
 		static let webFeedID = "webFeedID"
 		static let editedName = "editedName"
+		static let containerName = "containerName"
 	}
 
 	let url: String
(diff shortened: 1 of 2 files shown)


haikusw:
The renaming of functions is somewhat left over from my first attempt which used the old method for drags within the same window and the new method for drags between window.  So I had to have different versions of the functions for each strategy.  But then I removed the old method entirely because doing both involved too much duplicated code, made the code more complex, and made it too error prone to maintain (two spots to change any behavior).

So some names are longer than perhaps are needed, e.g., 

```
-	return localDragOperation(parentNode: parentNode)
+ 	return localDragFeedPasteboardOperation(parentNode: parentNode, draggedFeeds)
```

Didn't seem worth going back to modify that, but if you'd like me to I don't mind.

haikusw:
oh oh… just thought of another test case I never tried… (Really need to either change this code to facilitate automated tests or write a test plan for drag-n-drop).

haikusw:
Seems like they work:

In a single window:

1. Dragging multiple items from top level of two accounts to a third account -> works.
2. Dragging multiple items from a combination of top level and in a folder across two accounts to a third account -> works
3. Dragging multiple top level items from two accounts into a folder in one of the two accounts (has to copy the item from the other account and move the item from the same account, unless the option key is down and then needs to copy both).
4. Drag item from inside folders out to the top level where they are both from accounts different than the destination.
5. Drag item from inside folders out to the top level where one is from the same destination account and one is from a different account.


Do the same tests across two windows:
√ all seem to work.

<whew!>

vincode-io:
I've reviewed the code and tested it with iCloud and everything looks really well done to me. The only thing I would recommend we change before merging it is the function name thing. As mentioned in a previous comment, `localDragFeedPasteboardOperation` is more verbose than necessary. Let's switch it back to `localDragOperation`.

PR #3598, opened 2022-06-14
Add http and https to the list of CFBundleURLSchemes.

faisal:
If NetNewsWire says it is able to handle http and https URL schemes, Safari will list NetNewsWire as a target in Develop > Open Page With, which provides a trivial flow for subscribing to the page or site one is currently reading in Safari. I believe this functionality existed in NetNewsWire "classic," and this patch just brings the functionality back.

### DIFF
--- a/Mac/Resources/Info.plist
+++ b/Mac/Resources/Info.plist
@@ -37,6 +37,8 @@
 				<string>feed</string>
 				<string>feeds</string>
 				<string>x-netnewswire-feed</string>
+				<string>http</string>
+				<string>https</string>
 			</array>
 		</dict>
 	</array>


faisal:
To test:
1. Build + run NetNewsWire
2. Quit and relaunch Safari
3. With Safari's Developer menu enabled, look in Develop > Open Page With. Expected behavior: NetNewsWire (#.#) is in the list of browsers you can open the page in, with ## being the version of the built NetNewsWire (6.1 as of this writing)
4. Pick NetNewsWire (#.#) from Develop > Open Page With. Expected behavior: NetNewsWire comes to the foreground, and pops the subscription dialog with the page URL pre-filled. _Note_: NetNewsWire may ask for keychain access (e.g. to Feedly) after coming to the foreground and before popping the subscription dialog, but this request is incidental and can be denied.

PR #3599, opened 2022-06-14
Prefer expanding selected items upon right arrow key stroke on sidebar.

GetToSet:
This PR tweaks the right arrow keyboard shortcut on the sidebar to expand currently selected items if applicable rather than navigating to the timeline list. This results from a more predictable response to user actions.

https://user-images.githubusercontent.com/8158163/173612072-757fd66d-eb25-438d-89eb-88a81d756150.mov

### DIFF
--- a/Mac/MainWindow/MainWindowController.swift
+++ b/Mac/MainWindow/MainWindowController.swift
@@ -469,6 +469,12 @@ class MainWindowController : NSWindowController, NSUserInterfaceValidations {
 		currentTimelineViewController?.focus()
 	}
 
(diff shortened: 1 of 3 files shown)


vincode-io:
@brentsimmons this PR changes an established behavior in NetNewsWire. Is this something that was discussed in #work that I missed?

brentsimmons:
This isn’t something we’ve discussed. NetNewsWire’s arrow key behavior dates back to 2002, and is part of the character of NetNewsWire, and we don’t plan to change it.

(Note: cmd-rightArrow works to expand folders in the feeds list.)

Issue #3543, opened 2022-04-25
Window height grows 1px on Hide/Show Sidebar


matthewtodd:
It does indeed appear to be growing (changing h) and not shifting (changing y). The bottom edge of the window stays put.

This is NetNewsWire version 6.1 (6106) on latest Monterey, version 12.3.1 (21E258).

https://user-images.githubusercontent.com/5261/165170011-0fca113c-2824-46a4-9c69-950be29f62c5.mov

(Longtime fan here, thank you for all your hard work! This app is a gift.)

haikusw:
Replicated this.  Root cause is calling `bigSurOffsetFix()`.
I noticed that there's a console warning and set a breakpoint as directed and got the breakpoint inside this call and that makes it clear we have an issue (recursion):

`2022-06-15 22:51:41.455450-0700 NetNewsWire[31750:3818397] It's not legal to call -layoutSubtreeIfNeeded on a view which is already being laid out.  If you are implementing the view's -layout method, you can call -[super layout] instead. Break on void _NSDetectedLayoutRecursion(void) to debug.  This will be logged only once.  This may break in the future.
`

This is getting hit inside `bigSurOffsetFix()`; here's the backtrace:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00000001bf281f40 AppKit`_NSDetectedLayoutRecursion
    frame #1: 0x00000001bea50148 AppKit`-[NSWindow(NSConstraintBasedLayout) _layoutViewTree] + 148
    frame #2: 0x00000001bea4d1d0 AppKit`-[NSWindow _oldPlaceWindow:fromServer:] + 584
    frame #3: 0x00000001bea4c050 AppKit`-[NSWindow _setFrameCommon:display:fromServer:] + 2220
    frame #4: 0x0000000104ca2ad0 NetNewsWire`DetailWebView.bigSurOffsetFix(self=0x0000000106025550) at DetailWebView.swift:92:11
    frame #5: 0x0000000104ca2fc4 NetNewsWire`DetailWebView.setFrameSize(newSize=(width = 969, height = 849), self=0x0000000106025550) at DetailWebView.swift:62:4
    frame #6: 0x0000000104ca3020 NetNewsWire`@objc DetailWebView.setFrameSize(_:) at <compiler-generated>:0
[...]

PR #3603, opened 2022-06-16
Fix for #3543

haikusw:
Noticed that the window size wasn't returning the new value right after the code that sets its height to be old height -1 in one of the three times it's called during show or hide of the sidebar.  As a result, using the window size to calculate the size to restore the window to increments the original size instead of the -1 of that we thought we set.

IMPORTANT:  I'm not sure if this _breaks_ the fix that the `bigSurOffsetFix()` code was implementing because I can't replicate that bug on macOS 12.4 (I may just lack steps).

So someone who can replicate that should test this fix to see if that bug stays fixed or if this just effectively re-introduces it.

### DIFF
--- a/Mac/MainWindow/Detail/DetailWebView.swift
+++ b/Mac/MainWindow/Detail/DetailWebView.swift
@@ -90,7 +90,7 @@ final class DetailWebView: WKWebView {
 
 		frame.size = NSSize(width: window!.frame.width, height: window!.frame.height - 1)
 		window!.setFrame(frame, display: false)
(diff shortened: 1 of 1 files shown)


haikusw:
(note that this doesn't avoid the layout recursion Apple's log message warns about; why I'd prefer to just disable `bigSurOffsetFix()` in macOS 12 if the bug it fixes was fixed by Apple).

haikusw:
Sure.  Thanks for taking the time to review it.

Issue #662, opened 2019-05-19
Catch Up command


brentsimmons:
Catch Up…

Goes at the bottom of the Article menu. Displays a sheet:

Catch Up
Mark everything as read older than today in selected accounts:
[checkbox] On My Mac
[checkbox] FeedBin
etc.

[Cancel] [Catch Up]

If the [Catch Up] button is clicked, then it does what it says: marks everything as read older than today in the selected accounts.

Note: the [Catch Up] button is enabled when one or more accounts is enabled.

The list of accounts could be in a scrollable table view, with four lines (good guess?) showing at a time.

The checkbox defaults to unchecked for each account. This way there’s no accidental catching up.

brentsimmons:
Also note: this should be undoable, like the other mark-read commands.

vincode-io:
This is something I've wanted to have lately.  I think this issue is a good candidate for a beginner wanting to learn Mac programming a little better.  It is also something that someone wanting to contribute to iOS could do.

I envision this working quite differently on iOS.  I think it would go well in the context menu for Accounts.

vincode-io:
I'd like to propose an alternate UI that would be more flexible and work well on both iOS and macOS. (Since we don't have an Article menu on iOS.)

I'd like to add a context menu item in the context menu for the Sidebar. This menu item would be available for All Unread, Account, Folders, and Feeds. On macOS this menu should also be added to the Article menu. 

> Mark Older as Read in "Feed Name"

This context menu would expand a submenu with the following options:

* 1 day
* 2 days
* 3 days
* 1 week
* 2 weeks
* 1 month
* 1 year

(These are the same durations that BazQux uses for this function. I'm not married to them.)

The "Mark Older" item should only be available if there are unread articles in the selected item. I can't think of a way to efficiently restrict the duration submenu, so we would always have to show all options. 

The reason that I would like Catch Up to work more granularly than at the Account level is because different feeds have articles that age differently. For example a news feed has articles that are only relevant for a few days. But a developer blog has articles that can be relevant for years.

Another example is that I have a folder called Racing that has some very old, but still relevant articles about Road Racing. But if I get behind on my Twitter feeds in my Twitter folder, I would want to catch those up without marking my Racing articles as read.

I also think being able to select duration is useful. Maybe I have time to catch up on 2 or 3 days of Twitter posts and would like to do that instead of just saving 1 day.

brentsimmons:
[...]

PR #3668, opened 2022-08-28
Older Smart Feed

bryanculver:
While not a direct implementation of #662, this does provide a way to see all articles published before today.

To "Catch Up" a user can view the older feed and mark all as read.

The feed name "Old" is obviously open to change.

I have left some review comments asking for direction/feedback. This is my first Swift project so I might be missing some conventions.

Screen Shot 2022-08-28 at 09 19 31

Getting a build failure (duplicate copy) for Mac due to NewsFax theme.

### DIFF
--- a/ArticlesDatabase/Sources/ArticlesDatabase/ArticlesTable.swift
+++ b/ArticlesDatabase/Sources/ArticlesDatabase/ArticlesTable.swift
@@ -103,6 +103,16 @@ final class ArticlesTable: DatabaseTable {
 		fetchArticlesAsync({ self.fetchStarredArticles(webFeedIDs, limit, $0) }, completion)
 	}
 
+	// MARK: - Fetching Articles Between Dates
+
+	func fetchArticlesBetween(_ webFeedIDs: Set<String>, _ limit: Int?, since: Date? = nil, before: Date? = nil) throws -> Set<Article> {
+		return try fetchArticles{ self.fetchArticlesBetween(webFeedIDs, limit, $0, since: since, before: before) }
+	}
+
+	func fetchArticlesBetweenAsync(_ webFeedIDs: Set<String>, _ limit: Int?, _ completion: @escaping ArticleSetResultBlock, since: Date? = nil, before: Date? = nil) {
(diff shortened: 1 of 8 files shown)


ArticlesDatabase/Sources/ArticlesDatabase/ArticlesTable.swift
```diff
@@ -402,6 +412,52 @@ final class ArticlesTable: DatabaseTable {
 		}
 	}
 
+	func fetchUnreadCountBetween(webFeedIDs: Set<String>, since: Date? = nil, before: Date? = nil, completion: @escaping SingleUnreadCountCompletionBlock) {
+		// Get unread count between dates
```
bryanculver, line 416: Switches from positional parameters (borrowed implementation from `fetchUnreadCount`) to named parameters. `since` and `before` are optional and in the middle of the signature but because they are named parameters they don't pose issues with argument entry order.

ArticlesDatabase/Sources/ArticlesDatabase/ArticlesTable.swift
```diff
@@ -840,6 +896,34 @@ private extension ArticlesTable {
 		return fetchArticlesWithWhereClause(database, whereClause: whereClause, parameters: parameters)
 	}
 
+	func fetchArticlesBetween(_ webFeedIDs: Set<String>, _ limit: Int?, _ database: FMDatabase, since: Date? = nil, before: Date? = nil) -> Set<Article> {
```
bryanculver, line 899: As opposed to `fetchUnreadCountBetween`, I kept the positional-style arguments for all but the dates here, so they end up being mixed. Because `since` and `before` are optional, they get moved to the end of the signature.

Could potentially replace uses of `fetchArticlesSince` and `fetchArticles`.

vincode-io:
I'm closing this out since we will be implementing Catch Up in the 6.3 release cycle.

Thanks for your work on this @bryanculver. I hope you use this work as a jumping off point for helping us implement Catch Up.

PR #3670, opened 2022-09-03
Fix account add crash

winniequinn:
As reported in Ranchero-Software#3606, it was the case that attempting to add an account could result in a crash. This fixes the problem by retaining the `NSHostingController` that holds the `AddAccountsView` view in a manner similar to how 48138b1bb retained `accountsAddLocalWindowController`.


winniequinn:
While I am not sure this is the best possible fix—retaining the view controller indefinitely isn't ideal—it is the simplest fix that addresses the issue and it follows the pattern set in 48138b1bb4d54ac0e6b196a90b6ce76cbc02f8c3. It's unclear to me whether retaining the controller should be necessary at all or whether it merely works around a bug in SwiftUI.

This change is strictly better than what was there before, at least! I confirmed via manual testing that it fixes the crash and via profiling that no leaks are introduced.

vincode-io:
Thanks @winniequinn for finding this and figuring it out!

Issue #3512, opened 2022-03-30
miniflux greader errors


adamcstephens:
Hello. I'm using NNW iOS with the recently landed Miniflux greader support (mentioned in #2859). It works pretty well, but with a few issues/errors.

I spent some time digging in to the miniflux implementation, with the expectation that I'd need to fix the issues there. After investigating, they all seemed like they could be reported here first. Let me know if you disagree and I'd be happy to report them to miniflux or try and resolve them.

#### Unread state toggles back and forth in the NNW UI on each refresh

I verified miniflux is correctly returning the unread items every query to the greader API on the command line using `curl 'https://rss/reader/api/0/stream/items/ids?n=1000&output=json&s=user/-/state/com.google/starred' -H "Authorization: GoogleLogin auth=$GR_AUTH" | jq ".itemRefs | length"`

#### Adding new subscriptions causes a duplicate feed error

It appears that NNW is attempting to send two subscription requests.
First it sends a quickadd which succeeds
```
app_1  | [DEBUG] [HTTPS] 0.0.0.0 POST /reader/api/0/subscription/quickadd?quickadd=https://github.com/seashell/drago/releases.atom
````
Then it sends another `subscribe` to the subscription endpoint:
```
app_1  | [DEBUG] [HTTPS] 0.0.0.0 POST /reader/api/0/subscription/edit
app_1  | [INFO] [GoogleReader][/subscription/edit][ClientIP=0.0.0.0] Incoming Request for userID #2
app_1  | [DEBUG] Edit subscription map[T:[token] a:[user/-/label/Released] ac:[subscribe] s:[feed/134]]
```
Possibly miniflux could handle this, but it seems redundant to send two subscription requests.

#### Changing subscription labels fails

When trying to change the folder/label in NNW, I receive an error and nothing changes. It appears that NNW is attempting to remove the old label, prior to the adding the new label which doesn't appear to be supported by miniflux. https://github.com/miniflux/v2/blob/f6825c1c60a4e125b17f3cd27c26cccb2d520f35/googlereader/handler.go#L721-L729

The greader API seems to support the adding and removing in one API call, if I'm not mislead by https://github.com/mihaip/google-reader-api/blob/master/wiki/ApiSubscriptionEdit.wiki

```
app_1  | [DEBUG] [HTTPS] 0.0.0.0 POST /reader/api/0/subscription/edit
[...]

PR #3698, opened 2022-09-30
Fix iOS unread status indicator for ReaderAPI accounts

dave-atx:
Fix for issue #3512 where the unread status indicator toggles state on each refresh for some ReaderAPI accounts (at least miniflux accounts, perhaps others).

This PR changes `markAndFetchNew` to only notify for article IDs that actually change as opposed to all article IDs. This seems to match the original intent of the method and fixes the “status toggling” issue.

TBH, I’m unsure precisely _why_ this fixes the issue for ReaderAPI accounts — I believe there’s a race condition in the ReaderAPI code on iOS (but not macOS) that’s the root cause, and notifying only on new article statuses during a refresh happens to fix it.

### DIFF
--- a/Account/Sources/Account/Account.swift
+++ b/Account/Sources/Account/Account.swift
@@ -844,7 +844,7 @@ public final class Account: DisplayNameProvider, UnreadCountProvider, Container,
 		database.markAndFetchNew(articleIDs: articleIDs, statusKey: statusKey, flag: flag) { result in
 			switch result {
 			case .success(let newArticleStatusIDs):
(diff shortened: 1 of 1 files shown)


vincode-io:
I'm pretty sure that this change will prevent all accounts from updating the UI for article statues that were changed on a sync other than new articles. This function works correctly for all accounts currently including the ReaderAPI account type for services other than mini flux.

In fact, we shouldn't even be supplying the `newArticleStatusIDs` in the completion handler. That is obsolete code needed when we weren't syncing article content using CloudKit. I'll create an issue for myself to clean that up.

I'm going to close this pull request. If I misunderstood what you were trying to do and we are sure it hasn't broken any other syncing accounts, I'll reopen this and review it again.

Issue #3681, opened 2022-09-21
Feature request: Copy multiple article URLs


thefranke:
# Description
I'm using NetNewsWire regularly to pull for instance videos from multiple sites. The workflow is as follows:

* Right click *Copy Article URL* from RSS entry
* Paste into terminal with some scraping tool

I have to do this for each item individually. Ideally, there could be a *Copy Article URL* menu entry when selecting mutliple items in the feed, which will put a string of all article URLs separated by spaces into the clipboard.

# Example
I have selected 3 entries in the feed, pointing to https://url1.com, https://url2.com and https://url3.com. When selecting all three and using *Copy Article URL* the clipboard contains
https://url1.com https://url2.com https://url3.com

pdokas:
Check out third party clipboard managers like Pastebot or the ones built into tools like Alfred and Launchbar. They often have a feature that lets you concatenate instead of replace when copying or pasting from history like popping from a stack. Then you'd have this feature system wide instead of in select apps!

thefranke:
@pdokas Thanks for the suggestion, I'll give it a try. However, one thing this does not solve is that I'd still have to go through each individual item and select *Copy Article URL* instead of just selecting a bunch and getting the entire list with one click.

brentsimmons:
Reopening because this needs additional verification.

PR #3701, opened 2022-09-30
Support copying and opening multiple article URLs

Wevah:
Implements #3681 and #857.

Notes:

- If one of the selected articles has no URL when a copy is attempted, a one-time dialog is displayed letting the user know. (https://github.com/Ranchero-Software/NetNewsWire/pull/3701#issuecomment-1264320699)
- If more than 20 article URLs are selected to be opened in the browser, a confirmation alert is shown, and if more than 500 articles are selected, it becomes a no-op. (https://github.com/Ranchero-Software/NetNewsWire/pull/863#issuecomment-522701051)

### DIFF
--- a/Mac/Browser.swift
+++ b/Mac/Browser.swift
@@ -95,11 +95,11 @@ extension Browser {
 
 		if urlStrings.count > 20 {
 			let alert = NSAlert()
-			let messageFormat = NSLocalizedString("Are you sure you want to open %ld articles in your browser?", comment: "")
+			let messageFormat = NSLocalizedString("Are you sure you want to open %ld articles in your browser?", comment: "Open in Browser confirmation alert message format")
 			alert.messageText = String.localizedStringWithFormat(messageFormat, urlStrings.count)
-			let confirmButtonTitleFormat = NSLocalizedString("Open %ld Articles", comment: "")
+			let confirmButtonTitleFormat = NSLocalizedString("Open %ld Articles", comment: "Open URLs in Browser confirm button format")
 			alert.addButton(withTitle: String.localizedStringWithFormat(confirmButtonTitleFormat, urlStrings.count))
-			alert.addButton(withTitle: NSLocalizedString("Cancel", comment: ""))
+			alert.addButton(withTitle: NSLocalizedString("Cancel", comment: "Cancel button"))
 
 			if let window {
 				alert.beginSheetModal(for: window) { response in
--- a/Mac/Browser.swift
+++ b/Mac/Browser.swift
@@ -76,7 +76,13 @@ extension Browser {
 
 extension Browser {
 
-	static func open(_ urlStrings: [String], alertingInWindow window: NSWindow?, invertPreference: Bool = false) {
+	/// Open multiple pages in the default browser, warning if over a certain number of URLs are passed.
+	/// - Parameters:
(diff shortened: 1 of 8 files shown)


vincode-io:
Copy Article URL and Copy External URL work a little differently from each other. If there are any selected articles with an Article URL, then the Copy Article URL is available. That isn't the case with Copy External URL.

I'm honestly not sure which is better. Should we allow users to copy what they can if not all URL's are available? That might lead to users thinking that there is a bug and we dropped a URL (even though it never existed). On the other hand, if a URL is missing in the selection and we don't offer the option at all, they might also think there is a bug.

Maybe we allow the copy if one or more of the URL's is missing and display an alert when they copy explaining that X number of selected articles didn't have a URL available?

Wevah:
I didn't think I changed Copy External URL. Did I leave something in there about that?

Wevah:
Oh I see what you're saying, I think. I feel like if an article has no URL it's likely common in the feed, though I could be wrong.

vincode-io:
Smart Feeds and Folders can contain multiple Web Feeds which can differ on if they have URL's or not.

Wevah:
I think an alert should be fine if it includes the "don't tell me again" checkbox.

vincode-io:
Let's do it like we do our warning for the Full Screen stuff on the Article view.

https://github.com/Ranchero-Software/NetNewsWire/blob/f930eca9bba2f66707fcfe0798ac8907f12f989b/iOS/Article/ArticleViewController.swift#L299

Wevah:
> Smart Feeds and Folders can contain multiple Web Feeds which can differ on if they have URL's or not. 

Good point.

vincode-io:
As I mentioned in #work, let's fix how we open links in the Timeline context menu to work like it is supposed to.

Wevah:
Changing to a draft for now.

Mac/Browser.swift
```diff
@@ -73,3 +73,42 @@ extension Browser {
 			NSLocalizedString("Open in Browser in Background", comment: "Open in Browser in Background menu item title")
 	}
 }
+
+extension Browser {
+
+	static func open(_ urlStrings: [String], alertingInWindow window: NSWindow?, invertPreference: Bool = false) {
```
Wevah: This is to avoid lengthy code duplication.

Considered alternatives:

- Having this method on the `MainWindowController`. This would have the benefit of not needing the `window` parameter, but would require some sort-of-ugly `as? MainWindowController` stuff in the timeline view controller.
- Creating a `BrowserURLOpenable` protocol (with a better name), that provided this default implementation and required a `window` property (which `MainWindowController` already has). However, this `open` method should never be overridden.

One thing that could be changed is doing the check for the shift key in this method rather than passing in `invertPreferences`, since we're trying to unify behavior anyway.

vincode-io:
Did I click the wrong button regarding your question? It looks like my answer disappeared to me.

Basically, I said I like it the way it is. The alternatives weren't appealing. I agree about removing the `invertPreference` parameter from `Browser` and just checking for shift there.

vincode-io:
@Wevah Is there anything left to do with this? We're trying get the 6.2 release into user testing this week and just wondered if we could get this in there as well.

Wevah:
Let me rebase and push and I'll change this PR to non-WIP.

PR #3702, opened 2022-10-01
We don’t need to configure appearance

stuartbreckenridge:


vincode-io:
Is this true for all versions of iOS that we support now? I know we used to need to do the appearance config, but I don't know if we've moved past that now.

stuartbreckenridge:
Yes, I've had a look at physical iOS 16 devices and simulator iOS 15 devices.

PR #3706, opened 2022-10-01
Fix timeline, details views for proper toolbar scroll behavior

marioaguzman:
Both the details and timeline container views were not extending behind the toolbar and thus content would not show when scrolled behind it. Top anchors should anchor to the top of the super view rather than safe area guides and AppKit automatically adjusts the scroll insets for NSToolbar and optionally (if available) NSTitlebarAccessoryViewController. This allows for the divider to appear when content scrolls behind it and for content to "shine through" the translucency of the NSToolbar.

### DIFF
--- a/Mac/Base.lproj/MainWindow.storyboard
+++ b/Mac/Base.lproj/MainWindow.storyboard
@@ -3,6 +3,7 @@
     <dependencies>
         <deployment identifier="macosx"/>
         <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21225"/>
+        <capability name="NSView safe area layout guides" minToolsVersion="12.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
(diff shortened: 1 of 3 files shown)


marioaguzman:
Instead of the timeline rows and the details content getting "clipped" when scrolled under the toolbar, they now make the toolbar show the divider and the content now "shines through" the toolbar.
NNW_Toolbar_Fixed

brentsimmons:
When you switch to a different sidebar selection, does the translucent part update? It didn’t use to. Worth checking.

marioaguzman:
@brentsimmons - It does 😊 If you would like to see a video of it working, I have uploaded one here:
https://share.cleanshot.com/nqAO0p

You can see how the translucency behaves between the timeline column and the details column as well as how they reset once the user switches to another selection either in the timeline or the source sidebar.

vincode-io:
Thanks @marioaguzman!

marioaguzman:
You're most certainly welcome! Glad to help!

PR #3710, opened 2022-10-03
About Views

stuartbreckenridge:
- Tidied up iOS
- Adds refreshed About view to macOS 12+
- About.plist is now shared


vincode-io:
If you use two different hosting controllers, one for `AboutNetNewsWireView ` and one for `CreditsNetNewsWireView `, then you could swap them in `AboutWindowController `. Take a look at `PreferencesWindowController.switchToView` to see how to do this.

You wouldn't need to use the Notification then and you wouldn't need `AboutView` either.

It's a small change, but I think it would clean things up a little.

stuartbreckenridge:
@vincode-io This is now tidied up. The `AboutHostingController` manages the toolbar and segmented control, and will switch its `rootView` to About/Credits appropriately.

Issue #3723, opened 2022-10-26
Add a request limit/throttle for power users.


cocoonkid:
I am having the issue that NNW is pushing thousands of DNS requests at once when refreshing feeds.  This causes issues with pihole and probably many other services/firewalls/proxies.

In my case I am getting blocked for a minute :-)

One can clearly see the spikes caused by NNW on the scheduled refreshes.
![image](https://user-images.githubusercontent.com/43231640/198136134-184df23f-dff9-4760-9de0-e5ecd333997a.png)

It would be great to add an option for throttling or batching. At least when a certain threshold of subscribed feeds is reached.

I can see that this will cause issues most users won't immediately understand or cause blocks and false positive alerts for admins of networks etc.

Of course that might only be me.

Thank you for your consideration and the great software you provide for the community!

vincode-io:
We have discussed adding this feature in the past and decided against it. We want to keep NNW as easy to use as possible and a power user feature like this could be confusing to the average user.

Power users can configure the software that is throwing the errors so that they are more tolerant. 

The pihole limit of 1000 requests every 60 seconds is much too low for someone who clearly is trying to sync as many feeds as you have. Change the DNS `RATE_LIMIT` configuration and you should be fine.

https://docs.pi-hole.net/ftldns/configfile/

Issue #3732, opened 2022-11-06
Toolbar items not work on Text Only mode.


GetToSet:
Toolbar items not work on text-only mode. Items are all disabled and switched back to the normal mode once clicked.

<img width="1213" alt="2022-08-19 14 28 08" src="https://user-images.githubusercontent.com/8158163/200179781-33721000-ff10-4724-8eca-47b74ca5f8e1.png">

brentsimmons:
Reopening because this needs additional verification.

GetToSet:
I remembered a a PR to fix this issue was merged but looks like the commit is not at the current `main` branch.

brentsimmons:
Fixed in 7.1.3.

Issue #69, opened 2017-05-31
Article cleanup in database


brentsimmons:
Do on launch. (Maybe other times?)
When a feed has > 20 items, delete everything older than n days.

vincode-io:
I think this issue has been done and can probably be closed.

PR #3733, opened 2022-11-06
Allow toolbar buttons to work on text-only mode (#3732).

GetToSet:
This PR allows toolbar buttons to work on text-only mode (#3732).

`RSToolbarItem` should also be updated to accommodate this change. ([RSCore #69](https://github.com/Ranchero-Software/RSCore/pull/69))

### DIFF
--- a/Mac/MainWindow/MainWindowController.swift
+++ b/Mac/MainWindow/MainWindowController.swift
@@ -1386,19 +1386,25 @@ private extension MainWindowController {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


vincode-io:
@GetToSet If you use "Fixes #3723" in your commit message, GitHub will automatically close the issue when I merge it. That helps provide traceability between commits and the issue that the commit was for. It is very handy when trying to track down why a change was made.

PR #3742, opened 2022-11-12
iOS UI Settings Refresh

stuartbreckenridge:
Early re-work notes:
- A lot of `UIViewControllerRepresentable` shims over existing UIKit view controllers. These misbehave a little. They should be rewritten in SwiftUI.
- Settings no longer uses the Vibrant Table View Cell
- Add account button moved to the header instead of being a separate cell
- Changes to AppDefaults now trigger objectWillSend notifications.


stuartbreckenridge:
This is ready for testing.

Simulator Screen Shot - iPhone 14 Pro - 2022-12-04 at 21 12 55

stuartbreckenridge:
What's changed:

- Settings is now SwiftUI (as a result, NNW's vibrant cells are not used)
- List rows are in `SettingsViewRows`
- Iconography
- `AppDefaults` is now an `ObservableObject`
- Help/Info section of Settings has far less in it.
- Many of the existing UIKit controllers have SwiftUI _Representable_ wrappers, so they can't be retired just yet until they are re-written.

vincode-io:
Why are all the vars in SettingsViewRows capitalized?

The Manage Accounts view should have a title of Manage Accounts instead of Accounts. We should also be able to delete Accounts by swiping left on the table cells.

There is no obvious way to exit the Add Account view. After successfully adding an Account we should return to the Manage Accounts view without having to manually exit the Add Account view.

Manage Accounts and Manage Extensions are inconsistent with how we add items. Manage Extensions should also have the plus button and present a sheet to select which extension to add like it is in Manage Accounts. We should also be able to deactivate an Extension by swiping left.

Import/Export Subscriptions should be split into two separate table cells and we should use an Action Sheet to select which Account to use for each option. From the HIG, [Use an action sheet — not a menu — to provide choices related to an action.](https://developer.apple.com/design/human-interface-guidelines/components/presentation/action-sheets#ios-ipados)

stuartbreckenridge:
@vincode-io All addressed in 07d9c3d and ad66d70

iOS/Settings/Views/General/SettingsView.swift
```diff
@@ -0,0 +1,146 @@
...
+					viewModel.showImportExportError = true
+				}
+			}
+			.fileExporter(isPresented: $viewModel.showExportView, document: viewModel.exportDocument, contentType: OPMLDocument.writableContentTypes.first!, onCompletion: { result in
+				switch result {
+				case .success(_):
+					viewModel.showExportSuccess = true
+				case .failure(let error):
+					viewModel.importExportError = error
+					viewModel.showImportExportError = true
```
vincode-io, line 126: This doesn't look right.

vincode-io:
@stuartbreckenridge Did you delete all the now obsolete and unused classes?

stuartbreckenridge:
Not yet. Figured it should be tested then we can get rid.

iOS/Settings/Views/General/SettingsView.swift
stuartbreckenridge: In what way?

vincode-io:
@stuartbreckenridge Why did you close this pull request? Have you abandoned work on it?

stuartbreckenridge:
The work is in #3778

Issue #3748, opened 2022-11-14
Add Ability to Follow Mastodon Users


stuartbreckenridge:
How:

- Add a "Follow Mastodon User" option to the + button on the Master Feeds screen
- This will present UI that will ask the user for the username to follow in the format of <username>@<mastodon.domain>
- NNW will convert this to the RSS url: `https://<mastodon.domain>/users/ <username>.rss` and check to make sure it's valid
- if it's valid, it will add to the account / folder selected

stuartbreckenridge:
@brentsimmons Is this what you had in mind?

https://user-images.githubusercontent.com/7046652/203200831-da2743e7-94fe-41af-b31b-f2842ddbaca9.mov

alltom:
fyi, RSS feeds don't include boosts: https://github.com/mastodon/mastodon/issues/11290

Is there another protocol endpoint that would make NNW usable as a Mastodon client?

stuartbreckenridge:
The aim isn’t to make NetNewsWire a Mastodon client, but instead to make it easier to follow Mastodon RSS feeds.

alltom:
Yeah, but boosts are part of feeds. I got here because I noticed the RSS feeds of some of the people I followed were weirdly empty, and it turned out because most of their feeds were boosts.

NNW's Reddit support doesn't even use RSS, so I didn't think that would be a blocker here…

vincode-io:
@alltom you are correct. NNW uses the Reddit API, not RSS for its support. I don’t think we can rule out using ActivityPub for a Mastodon Extension in the future.

PR #3755, opened 2022-11-22
Adds the ability to add Mastodon feeds to iOS

stuartbreckenridge:
This adds the ability to add Mastodon feeds to iOS (related to #3748)

If we are happy with this implementation, I'll work on the macOS equivalent.


vincode-io:
Why is this UI so very different than the other 4 options we have to add items?

stuartbreckenridge:
Visually, it's more inline with Add Account UI. 

It's doing more than standard Add Feed UI, given the federated nature of Mastodon. 

We need username and instance. To get that we could just ask for the full URL from the user and append `.rss` and see if it's correct. But I felt splitting out username and instance was a better approach. 

For following tags, the UI doesn't give an option for the instance on which to follow the tags (which I actually think I need to correct).

vincode-io:
It does look like an Add Account UI, but nothing like any thing else under the plus button. None of the other items there have logos or banners or explanation blocks at the bottom.

The Add Reddit Feed and Add Twitter Feed do just as much as the Add Mastodon Feed, if not more. They are done in completely different ways though.

Shouldn't we try to make sure the NetNewsWire UI is consistent for both usability reasons and aesthetic ones?

stuartbreckenridge:
The other option I can think of would be to utilise the Mastodon [Search API](https://docs.joinmastodon.org/methods/search/). 

If we used Mastodon.social the lookup would be limited to Mastodon.social and its _federation_.

vincode-io:
I don't have a problem with username/tag and instance being separate fields.

The username vs. tag should be a table that pushes the next screen for the user like the Twitter and Reddit add options. We should also use the same mechanism to select the Account/Folder as all the other options currently do.

I understand if you think the Mastodon option is done better. It doesn't really matter though unless we commit to changing all the other Add options as well. Consistency in app design is very important. We don't want NetNewsWire to look like a bunch of guys that don't talk to each other. We want it to look like it was made by a cohesive team, right?

stuartbreckenridge:
I get what you mean now that I've been noodling in these views for a few days. I'll come back to this once all the Settings stuff is squared away.

stuartbreckenridge:
Will rework this!

Issue #3774, opened 2022-12-14
[iOS] Re-write "Administrative" Views in SwiftUI


stuartbreckenridge:
As a result of #3742, some of these controllers are currently wrapped in `Repesentable`s. While this works, it relies on hooking into `UIStoryboard`, which, in turn, means more work down the road (e.g., localisation of `.xib`s, maintaining UIKit assets to support SwiftUI code, etc.). 

### Settings 

- [ ] `AddAccountViewController`
- [ ] `AddExtensionPointViewController`
- [ ] `ArticleThemesTableViewController`
- [ ] `EnableExtensionPointViewController`
- [ ] `TimelineCustomizerViewController`
- [ ] `TimelinePreviewTableViewController`

### Inspector

- [ ] `AccountInspectorViewController`
- [ ] `ExtensionInspectorViewController`
- [ ] `WebFeedInspectorViewController`

### Account

- [ ] `LocalAccountViewController`
- [ ] `CloudKitAccountViewController`
- [ ] `FeedbinAccountViewController`
- [ ] `NewsBlurAccountViewController`
- [ ] `ReaderAPIAccountViewController`

Issue #3738, opened 2022-11-12
iOS Settings Revamp


stuartbreckenridge:
In no particular order:
- [ ] Rewrite in SwiftUI
- [ ] Similar to iOS Settings, add some iconography to visually distinguish sections 
- [ ] Judicious use of context menus (e.g., for changing dark/light mode)

vincode-io:
You should only use MenuPickerStyle if you have 5 or more items according to Apple's documentation. It is important to use the correct PickerStyle to make the Settings as easy to use and discoverable as possible.

stuartbreckenridge:
@vincode-io Agreed. I just love the menu animation!

PR #3778, opened 2022-12-18
Settings/Inspectors/Account to SwiftUI

stuartbreckenridge:
Lots of changes related to #3738 & #3774:

- All Inspectors are now SwiftUI and localised
- All Add Account/Credentials controllers are now SwiftUI and localised
- All Settings views are SwiftUI and localised

Everything is behaving pretty much as it did under `UIKit`.  

### New

- Settings revamped with iconography and rejigged ordering.
- The Account Management list (which is new) is split into Active/Inactive accounts and includes swipe actions to toggle the active and inactive states. Swipe actions are available on Extensions as well. 
- Localisation is not handled in the PR (it has moved to #3811)
- SwiftUI's `fileExporter` and `fileImporter` are being used for OPML files and `.nnwtheme` packages
- Unused controllers and XIBs have been deleted.
- The SwiftUI Timeline customiser uses a _cheat_ view that is built to look like the `UIKit` cell. 
- The Article Themes list now credits the theme creator and separates Ranchero provided themes from third-party themes.
- Several existing classes have been made `ObservableObjects` so that they can fire `objectWillChange` notifications to `@StateObject`s.

### Fixes

- Alerts/Action Sheets now consistently tint the destructive actions (e.g., remove account) in red
- Account explainers are now consistently displayed as footers and not rows. 
- Timeline Customiser now works in landscape (it can be scrolled when the user has selected more rows than are visible)

### Known Issue

- Rarely, the inspector views will have a larger offset from the top than they should have (hence the `.edgesIgnoreSafeArea(.bottom)` modifier usage). Why this happens is not clear.
 

Some previews: 

Sunday, 18 Dec 2022 19:23:38

Sunday, 18 Dec 2022 19:26:09

Sunday, 18 Dec 2022 19:22:37

Thursday, 22 Dec 2022 06:51:52


iOS/Settings/Account and Extensions/Extensions/EnableExtensionViewModel.swift
```diff
@@ -0,0 +1,182 @@
...
+	
+	public var rootViewController: UIViewController? {
+		var currentKeyWindow: UIWindow? {
+			UIApplication.shared.connectedScenes
+				.filter { $0.activationState == .foregroundActive }
+				.map { $0 as? UIWindowScene }
+				.compactMap { $0 }
+				.first?.windows
+				.filter { $0.isKeyWindow }
+				.first
```
vincode-io, line 172: Maybe this should be added to `UIWindow+RSCore.swift`.

iOS/UIKit Extensions/UIViewController-Extensions.swift
```diff
@@ -13,9 +13,7 @@ import Account
 extension UIViewController {
 	
 	func presentError(_ error: Error, dismiss: (() -> Void)? = nil) {
-		if let accountError = error as? AccountError, accountError.isCredentialsError {
-			presentAccountError(accountError, dismiss: dismiss)
-		} else if let decodingError = error as? DecodingError {
```
vincode-io, line 18: Why was this removed?

vincode-io:
Why does adding an Extension push a navigation stack instead of showing a sheet like happens when adding an Account?

When you try to add a Feedbin account the sheet that appears just says "Feedbin" on it. How did you test that the Feedbin sheet works correctly? You did test all the different Add Sheets you created to make sure they work right?

We might want to name the *AddAccountView views something else. They are used both for Adding and Updating Credentials. It through me off for a bit when I saw them being used in `AccountInspectorView`.

When the Setting are opened in portrait on an iPad, there is a lot of whitespace at the bottom of the view. Is there anything we can do about this?

Any idea why some of these footer explainers are breaking the line too soon?

Screen Shot 2022-12-20 at 4 58 26 PM

I'm getting an error when trying to add a Theme that didn't occur in the old code. This is for a Theme that is in my Documents folder on iCloud Drive.

Screen Shot 2022-12-20 at 5 11 28 PM


The Timeline slider should revert back to the old design.

vincode-io:
Was that last commit comment an accident? Could you please amend it so that it describes what was changed?

stuartbreckenridge:
> Why does adding an Extension push a navigation stack instead of showing a sheet like happens when adding an Account?
- Changed to `.sheet` in 1a1f86f
-
> When you try to add a Feedbin account the sheet that appears just says "Feedbin" on it. How did you test that the Feedbin sheet works correctly? You did test all the different Add Sheets you created to make sure they work right?
- There was a missing case for Feedbin, which has been added. The remainder, inc. Feedly were tested. Same for the Reddit extension. Fixed in 1a1f86f

> We might want to name the *AddAccountView views something else. They are used both for Adding and Updating Credentials. It through me off for a bit when I saw them being used in AccountInspectorView.
- `*AccountManagementView` ?

> When the Setting are opened in portrait on an iPad, there is a lot of whitespace at the bottom of the view. Is there anything we can do about this?
- Not that I am aware of.

> Any idea why some of these footer explainers are breaking the line too soon?
- It's the same in `UIKit` on iPhone 12 mini as far as I can tell from the simulator. I don't think there's much we can do.

stuartbreckenridge:
> Was that last commit comment an accident? Could you please amend it so that it describes what was changed?

Not a mistake. The changes were in the commit description.

vincode-io:
My mistake. I didn't see that it was a multi line commit.

But you are doing something that I had a hard time breaking myself of which is forgetting to always start your commit with a verb. https://github.com/Ranchero-Software/NetNewsWire/blob/main/Technotes/CodingGuidelines.md#version-control

stuartbreckenridge:
> The Timeline slider should revert back to the old design.
- Restored in dee9461

iOS/UIKit Extensions/UIViewController-Extensions.swift
stuartbreckenridge: There are no `UIViewController`s left that make use of that code path.

iOS/Settings/Account and Extensions/Accounts/AccountsManagementView.swift
```diff
@@ -0,0 +1,155 @@
...
+			self?.refreshAccounts()
+		}
+		
+		NotificationCenter.default.addObserver(forName: .UserDidDeleteAccount, object: nil, queue: .main) { [weak self] _ in
+			self?.refreshAccounts()
+		}
+		
+		NotificationCenter.default.addObserver(forName: .DisplayNameDidChange, object: nil, queue: .main) { [weak self] _ in
+			self?.refreshAccounts()
+		}
```
Mackarous, line 39: I feel like I'm coming out of nowhere contributing here but every like 6-12  months I find myself coming back to this project and attempting to put some time into it

Anyways, this `NotificationCenter` stuff could use the new [`assign(to:)`](https://developer.apple.com/documentation/combine/fail/assign(to:)) API like so if you're inclined:

```swift
		let changes = Publishers.Merge4(
			NotificationCenter.default.publisher(for: .AccountStateDidChange),
			NotificationCenter.default.publisher(for: .UserDidAddAccount),
			NotificationCenter.default.publisher(for: .UserDidDeleteAccount),
			NotificationCenter.default.publisher(for: .DisplayNameDidChange)
		)
		
		changes
			.map { _ in AccountManager.shared.sortedActiveAccounts }
			.assign(to: &$sortedActiveAccounts)
		
		changes
			.map { _ in AccountManager.shared.sortedAccounts.filter { !$0.isActive} }
			.assign(to: &$sortedInactiveAccounts)
```

Mackarous: In fact you could even do away with the initial call to `refreshAccounts()` by doing the following also:
```swift
	init() {
		let changes = Publishers.Merge4(
			NotificationCenter.default.publisher(for: .AccountStateDidChange),
			NotificationCenter.default.publisher(for: .UserDidAddAccount),
			NotificationCenter.default.publisher(for: .UserDidDeleteAccount),
			NotificationCenter.default.publisher(for: .DisplayNameDidChange)
		).share()
		
		changes
			.map { _ in AccountManager.shared.sortedActiveAccounts }
			.prepend(AccountManager.shared.sortedActiveAccounts)
			.assign(to: &$sortedActiveAccounts)
		
		changes
			.map { _ in AccountManager.shared.sortedAccounts.filter { !$0.isActive} }
			.prepend(AccountManager.shared.sortedAccounts.filter { !$0.isActive})
			.assign(to: &$sortedInactiveAccounts)
	}
```

iOS/Settings/Appearance/ArticleThemeManagerView.swift
```diff
@@ -0,0 +1,182 @@
...
+	@StateObject private var themeManager = ArticleThemesManager.shared
+	@State private var showDeleteConfirmation: (String, Bool) = ("", false)
+	@State private var showImportThemeView: Bool = false
+	@State private var showImportConfirmationAlert: (ArticleTheme?, Bool) = (nil, false)
+	@State private var showImportErrorAlert: (Error?, Bool) = (nil, false)
+	@State private var showImportSuccessAlert: Bool = false
+	
+	var body: some View {
+		Form {
+			Section(header: Text("Built-in Themes", comment: "Section header for installed themes"), footer: Text("These themes cannot be deleted.", comment: "Section footer for installed themes.")) {
```
Mackarous, line 22: This `Section(header:` call is [deprecated in iOS 16.4](https://developer.apple.com/documentation/swiftui/section/init(header:content:))

They suggest to use [`Section(content:, header:)`](https://developer.apple.com/documentation/swiftui/section/init(content:header:)-29e2l) instead

iOS/Settings/Appearance/ArticleThemeManagerView.swift
```diff
@@ -0,0 +1,182 @@
...
+struct ArticleThemeManagerView: View {
+    
+	@StateObject private var themeManager = ArticleThemesManager.shared
+	@State private var showDeleteConfirmation: (String, Bool) = ("", false)
+	@State private var showImportThemeView: Bool = false
+	@State private var showImportConfirmationAlert: (ArticleTheme?, Bool) = (nil, false)
+	@State private var showImportErrorAlert: (Error?, Bool) = (nil, false)
+	@State private var showImportSuccessAlert: Bool = false
+	
+	var body: some View {
```
Mackarous, line 20: I was playing around with a preview for the theme manager
Let me know if you want to connect over it and put it in:
Simulator Screen Shot - iPhone 14 - 2023-02-22 at 15 46 33

iOS/Settings/Account and Extensions/Accounts/AccountsManagementView.swift
stuartbreckenridge: > I feel like I'm coming out of nowhere contributing here but every like 6-12 months I find myself coming back to this project and attempting to put some time into it
> 
> Anyways, this `NotificationCenter` stuff could use the new [`assign(to:)`](https://developer.apple.com/documentation/combine/fail/assign(to:)) API like so if you're inclined:
> 
> ```swift
> 		let changes = Publishers.Merge4(
> 			NotificationCenter.default.publisher(for: .AccountStateDidChange),
> 			NotificationCenter.default.publisher(for: .UserDidAddAccount),
> 			NotificationCenter.default.publisher(for: .UserDidDeleteAccount),
> 			NotificationCenter.default.publisher(for: .DisplayNameDidChange)
> 		).share()
> 		
> 		changes
> 			.map { _ in AccountManager.shared.sortedActiveAccounts }
> 			.assign(to: &$sortedActiveAccounts)
> 		
> 		changes
> 			.map { _ in AccountManager.shared.sortedAccounts.filter { !$0.isActive} }
> 			.assign(to: &$sortedInactiveAccounts)
> ```

We don't use the Combine framework in that way anywhere else in the app. It would be inconsistent to use it here.

Mackarous: No worries, it may be inconsistent but so is SwiftUI, new shift could be good for new paradigm. Sometimes it makes sense to use the new APIs. 

In this specific instance you actually don't need to concern about reference cycles, reducing room for error.

stuartbreckenridge: I think the closest _new_ thing to use would `AsyncSequence`. For example, in `MasterFeedController`, there are eight `NotificationCenter.default.addObserver`s in `viewDidLoad`. In `UIKit`, they'd be replaced:

```swift

private var notificationsTask: Task<(), Never>?

override func viewDidLoad() {

    // NotificationCenter.default.addObserver x8 removed
    
    notificationsTask = Task {
        await notificationHandler()
    }

}


func notificationHandler() async {
    for await notification in NotificationCenter.default.notifications(named: .UnreadCountDidChange) {
        unreadCountDidChange(notification)
    }
    
    // ... and the rest
}
```

You could then get rid of `@objc` from the method signatures. 

In `SwiftUI`, the above is simpler, as you don't need to worry about retain cycles when views are removed. The `notificationHander` function can be added to a `.task` modifier.

Interesting architectural discussion for @brentsimmons  / @vincode-io.

brentsimmons:
@stuartbreckenridge Looks like there are some conflicts to resolve — let me know when those are resolved and I’ll review this PR.

stuartbreckenridge:
@brentsimmons That should be it: 
- SettingsViewController is no longer present in this PR, so it's fair to remove.
- The SwiftUI PR has @vincode-io's work related to Mark as Read on Scroll added.

brentsimmons:
@stuartbreckenridge would you also update this one to fix the conflict, so I can get it merged?

stuartbreckenridge:
@brentsimmons Conflicts resolved. Note this should be merged _before_ the internationalization commit.

PR #3818, opened 2023-01-19
Ignores files in __MACOSX folder

### LINKED ISSUE #3817: Ignore __MACOSX folder when searching for unzipped theme files
In edge cases, the search will try to use the `.nnwtheme` in this folder, which is empty.

stuartbreckenridge:
Fixes #3817


gruber:
“Contains” is probably fine, but is there a method for StartsWith? And maybe starts with “__MACOSX/“ with the slash?

PR #3820, opened 2023-01-20
New RSS Icon (in the style of SF Symbols)

mantia:


mantia:
Updated the icon again, now it should be correctly sized and positioned for Safari’s toolbar.

danielpunkass:
I confirmed that with the updated icon on this PR the built app extension produces an icon like this:

image

I think we should accept and merge this at @brentsimmons or @vincode-io 's convenience. Thanks @mantia for the idea to update the icon, and for sticking with the many revisions it took to defeat Safari's default image sizing/placement.

PR #3911, opened 2023-03-13
Fixes #3910

### LINKED ISSUE #3910: UITableView laid out while not in hierarchy
Just caught this in console:

2023-03-12 15:32:57.160381-0700 NetNewsWire[23078:800895] [TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. Table view: <UITableView: 0x12402d600; frame = (0 0; 390 844); clipsToBounds = YES; autoresize = W+H; gestureRecognizers = <NSArray: 0x600000fcfe40>; backgroundColor = <UIDynamicSystemColor: 0x60000140fb80; name = tableBackgroundColor>; layer = <CALayer: 0x6000001f8000>; contentOffset: {0, 221.66666666666666}; contentSize: {390, 61235.333333333358}; adjustedContentInset: {91, 0, 83, 0}; dataSource: <_TtGC11NetNewsWire24MasterTimelineDataSourceSiV8Articles7Article_: 0x60000037f080>>

stuartbreckenridge:
`UITableViewAlertForLayoutOutsideViewHierarchy` will no longer fire as a check has been added to make sure that the tableView’s `window` property is not nil—i.e., it's in the view hierarchy—before performing batch updates.

### DIFF
--- a/iOS/MasterFeed/MasterFeedViewController.swift
+++ b/iOS/MasterFeed/MasterFeedViewController.swift
@@ -564,6 +564,8 @@ class MasterFeedViewController: UITableViewController, UndoableCommandRunner, Ma
 			return
 		}
 		
+		if tableView.window == nil { return }
+		
 		tableView.performBatchUpdates {
 			if let deletes = changes.deletes, !deletes.isEmpty {
 				tableView.deleteSections(IndexSet(deletes), with: .middle)
--- a/iOS/MasterFeed/MasterFeedViewController.swift
+++ b/iOS/MasterFeed/MasterFeedViewController.swift
@@ -564,7 +564,10 @@ class MasterFeedViewController: UITableViewController, UndoableCommandRunner, Ma
 			return
 		}
 		
-		if tableView.window == nil { return }
+		if tableView.window == nil {
+			completion?()
+			return
+		}
 		
 		tableView.performBatchUpdates {
 			if let deletes = changes.deletes, !deletes.isEmpty {


brentsimmons:
I think the bug mentioned TimelineDataSource — I think this is happening in the timeline rather than in the Feeds view.

iOS/MasterFeed/MasterFeedViewController.swift
```diff
@@ -564,6 +564,8 @@ class MasterFeedViewController: UITableViewController, UndoableCommandRunner, Ma
 			return
 		}
 		
+		if tableView.window == nil { return }
```
brentsimmons, line 567: We still should call the completion — and we probably do want to call `reloadData` too, which I think is smart enough not to do layout when it’s not actually warranted. (I think.)

brentsimmons (changes requested):

stuartbreckenridge:
Based on the code tree, it appears it is the correct place to fix it:

Screenshot 2023-03-13 at 13 29 24

iOS/MasterFeed/MasterFeedViewController.swift
stuartbreckenridge: From testing, we can call the completion safely, but the `reloadData` opens up another avenue to hit this bug — it appears to trigger an unreadCountDidChange notification, which trickles down to a layout of the cell

Screenshot 2023-03-13 at 13 44 49

brentsimmons (approved):

Issue #3734, opened 2022-11-06
Mark All as Read Behavior


vincode-io:
## The Issue

I don't think the Mark All as Read function should mark items that I have begun to read and then marked as unread to read when used. My reasoning is that because I have explicitly told NetNewsWire that I want that article marked as unread, NetNewsWire shouldn't automatically mark it as read when the Mark All as Read function is used.

## My Reasoning

The main reason that I will mark an article as unread is that I don't have the time or energy to read an article that I may have opened which marked it as unread. I could star the article, but that isn't what I use starred status for. 

To me starred is for articles that I want more permanence in how long it is saved. For example, I have [Supporting async/await as a module owner](https://www.swiftindepth.com/articles/async-await-modules) starred because I think I might need it to migrate some of the RS project code to async/await. That might not happen soon and I want to be able to reference this article at anytime in the future.

That is different from an article that I want to read soon, but don't mind losing if I never get to it. I guess I could use a Read Later service, but I don't. I figure that we have Mark as Unread for a reason and that reason is because I want to see it in the Timeline the next time I pull that Timeline up.

Because I use Mark as Unread (for what it intended, I believe) I have to be very careful when using Mark All as Read, Mark Above as Read, or Mark Below as Read. I have to keep it in my head that as long as I am in a Timeline that I have done a Mark as Unread, that I shouldn't use the bulk marking functions.

I didn't even know I was doing this until we were working on the design for [Mark Read on Scroll](https://github.com/Ranchero-Software/NetNewsWire/issues/3552), but I have been thinking about the behavior of Mark All as Read ever since.

## The Consequences

I realize that there is a lot of inertia to changing the behavior of Mark All as Read. Rightfully so. It has literally worked the same way for decades. Will making this change shock or upset NetNewsWire users though?

If this change were made, I believe we will have positive user feedback, not negative when users first encounter it. When users first see that Mark All as Read left something in the Timeline as unread, they will most likely check the Timeline and recognize the article that they had marked as unread. Personally, I would be relieved that Mark All as Read hadn't undone my Mark as Unread command and that the article still appeared in my Timeline.

brentsimmons:
Over the years I’ve heard from many people that they use mark-unread this way — it’s how they maintain their to-read list.

One of the planned features for NetNewsWire 5.0 that didn’t make it was to have a To Read folder, much like the Starred folder, where you put articles that you want to read. The idea is that it would be short-term storage, not permanent like Starred.

[...]

Issue #1844, opened 2020-03-01
RFE: Mark read on scroll in timeline


vincode-io:

brentsimmons:
How would this behave exactly? For instance, say I tap on a feed and the timeline opens. Is everything visible in the timeline marked as read immediately? Or do I have to scroll a little first? Or…?

vincode-io:
This was requested more than a couple times. I know at least once or twice in test flight feedback and I think in the feedback slack channel as well. 

I believe that the goal here is to make the timeline work more like a infinite scroll social media timeline. Once you scroll past a timeline item it doesn’t come back again. 

Since we always scroll up, an item would get marked as read after it was visible and then gets scrolled out of site through the top of the scroll area.

Bunn:
@brentsimmons @vincode-io I played around with an implementation of it some time ago and came to this approach:
https://www.dropbox.com/s/ftob2f62somw9ir/autoread.mov?dl=0

One problem that this might cause is that the last items on the table would never be marked as read during a scroll and the user would need to manually mark as read anyways, breaking out the seamless flow that this feature provides.

One possible way to solve this is to change the table contentInset a little bit to allow the user to scroll more than necessary but it might not be visually pleasant, looking like a tableview bug to scroll and see no new content.

brentsimmons:
What other app does this well? I’d like to look at prior art.

vincode-io:
Unread has this feature.

Bunn:
Feedly

scapix:
Feedly - "Mark all as read" button at bottom of the list
Newsify - pull up to mark all as read when at the bottom (similar to how refreshing stuff works in most up, just in opposite direction)

I prefer Feedly implementation as in Newsify you are able to trigger it accidentaly when trying to scrolldown and not noticing you are at the end.

abonham:
I second the Feedly approach, it's the only thing that's stopping me switching full time. It's frustrating when I've scrolled past a number of things I'm not interested and forget to mark all above as read if I need to do something else before getting to FeedZero

vincode-io:
We have a Mark All as Read button at the bottom of the Timeline at all times. Another one like Feedly has would be redundant for us. 

I think all we need to do is mark the article as read after it leaves the viewable area.

scapix:
[...]

PR #3913, opened 2023-03-13
Mark as Read on Scroll

vincode-io:
This pull request implements https://github.com/Ranchero-Software/NetNewsWire/issues/1844.

I started with @everhardt's original work on this issue. That has some problems so I rewrote the everything except the Settings code. My new code had problems too, so I rewrote that. That code caused hitching when scrolling the Timeline, so that got rewrote as well. This implementation works well and there is no noticeable degradation in performance when using Mark as Read on Scroll, even on my 6 year old iPad.

Nothing was changed in NetNewsWire's UI except in the Settings to enable/disable this feature.

The sticking point in implementing this feature previously was what to do when we reached the bottom of the Timeline and wanted to mark the rest of the articles as read. To do this we that we would either have had to add a different Mark All as Read control or change how the current Mark All as Read control worked. Because [we changed how the current Mark All as Read works](https://github.com/Ranchero-Software/NetNewsWire/issues/3734), this is no longer an obstacle to implementation.


Mac/MainWindow/Timeline/TimelineViewController.swift
```diff
@@ -124,6 +124,7 @@ final class TimelineViewController: NSViewController, UndoableCommandRunner, Unr
 			}
 
 			directlyMarkedAsUnreadArticles = Set<Article>()
+			lastVerticlePosition = 0
```
brentsimmons, line 127: Nit — can be dealt with in future PR — `lastVerticlePosition` should be `lastVerticalPosition`

brentsimmons (approved):

vincode-io:
Yeah, I can't spell. I'll submit something to fix it in a second.

PR #3914, opened 2023-03-14
Fix lastVerticalPosition variable spelling

vincode-io:


brentsimmons (approved):
Was there also a case in `MasterTimelineViewController`?

vincode-io:
Of course. Sigh... I should have thought of that.

PR #3991, opened 2023-05-28
Fix localization keys for local account names in Account package

J-rg:


brentsimmons:
@stuartbreckenridge As the localization expert, would you review this PR?

stuartbreckenridge:
Account Names are currently localized via the main app's `.strings` files. As it stands, replacing the keys in the Account package will not do anything. 

```swift
func localizedAccountName() -> String {
		
		switch self {
		case .onMyMac:
			let defaultName: String
			#if os(macOS)
			defaultName = NSLocalizedString("account.name.mac", comment: "On My Mac")
			#else
			if UIDevice.current.userInterfaceIdiom == .pad {
				defaultName = NSLocalizedString("account.name.ipad", comment: "On My iPad")
			} else {
				defaultName = NSLocalizedString("account.name.iphone", comment: "On My iPhone")
			}
			#endif
			return defaultName
		
		/* The below account names are not localized as they are product names. */
			
		case .bazQux:
			return "BazQux"
		case .cloudKit:
			return "iCloud"
		case .feedbin:
			return "Feedbin"
		case .feedly:
			return "Feedly"
		case .freshRSS:
			return "FreshRSS"
		case .inoreader:
			return "Inoreader"
		case .newsBlur:
			return "NewsBlur"
		case .theOldReader:
			return "The Old Reader"
		}
	}
```

brentsimmons:
@stuartbreckenridge Does that mean this PR should be closed?

stuartbreckenridge:
Yes. For the moment, they are localized in the app.

stuartbreckenridge:
If we take this code from `AddAccountListView` on iOS:

```swift
Text(verbatim: AccountType.onMyMac.localizedAccountName())
```

The `localizedAccountName()` uses the app's `Localizable.strings` file: 

```
/* On My iPad */
"account.name.ipad" = "On My iPad";

/* On My iPhone */
"account.name.iphone" = "On My iPhone";

/* On My Mac */
"account.name.mac" = "On My Mac";
```

stuartbreckenridge:
Note that there is a `TODO` to move `localizedAccountName` to the `Account` package.

stuartbreckenridge:
That's my point @J-rg — we do not need to localize the account names in the Account package just now because it's handled in the main app. 

If that changes in the future, this PR is 100% correct in that we'd need to re-key the strings in code, but we'd also need to add translations in the Account package's `Localizable.strings`.

stuartbreckenridge:
@brentsimmons Let's approve this. @J-rg has valid points and this is an additive PR. I will do some follow-up work to move Account name-related localizations from the main app to the Account package.

PR #4005, opened 2023-05-31
Remove Deprecation Warnings

### LINKED ISSUE #4004: Fix deprecation warnings

stuartbreckenridge:
Current state with clean builds:

- iOS warnings — 0
- macOS warnings — 1

The macOS warning is related to the Sharing context menu, and I'm suggesting we leave it:

```
'sharingServices(forItems:)' was deprecated in macOS 13.0: Use -[NSSharingServicePicker standardShareMenuItem] instead.
```

The crucial updates:

- iOS now targets iOS 16
- `OPML` is now an importable UTI type
- `allowedFileTypes` is removed in favour of `allowedContentTypes`
- `NavigationLink` replaced with `NavigationStack`
- `InteractiveLabel` had a deprecation warning, but the class wasn't used and the file has been removed.

This will close #4004


Mac/MainWindow/OPML/ExportOPMLWindowController.swift
```diff
@@ -75,7 +75,7 @@ import Account
 	func exportOPML(account: Account) {
 		
 		let panel = NSSavePanel()
-		panel.allowedFileTypes = ["opml"]
+		panel.allowedContentTypes = [.xml]
```
brentsimmons, line 78: Should this be `.opml` instead of `.xml`?

iOS/Settings/General/SettingsRows.swift
```diff
@@ -49,15 +49,24 @@ struct SettingsRow {
...
+//		NavigationLink(destination: AccountsManagementView()) {
+//			Label {
+//				Text("button.title.manage-accounts", comment: "Manage Accounts")
+//			} icon: {
+//				Image("app.account")
+//					.resizable()
+//					.frame(width: 25.0, height: 25.0)
+//					.clipShape(RoundedRectangle(cornerRadius: 6))
+//			}
+//		}
```
brentsimmons, line 61: Should we delete this code instead of commenting it out?

brentsimmons (changes requested):

Mac/MainWindow/OPML/ExportOPMLWindowController.swift
stuartbreckenridge: I've addressed this in 7a008db — `allowedContentTypes` is `.opml` and, to ensure it is used, the Save panel doesn't provide an extension in `nameFieldStringValue` as the system provides it from the `allowedContentTypes` array.

iOS/Settings/General/SettingsRows.swift
stuartbreckenridge: I've fixed this in 17c4252.

brentsimmons (approved):

PR #4007, opened 2023-05-31
Enable Base localization for remaining XIB files

J-rg:


stuartbreckenridge:
✅ Good to go. I missed these ones.

PR #4025, opened 2023-06-10
Removes Reddit Integration on the Mac

stuartbreckenridge:
- Removes Reddit FeedProvider
- Removes Reddit related UI, Assets, and Extension Preferences pane
- Adds deprecation alert

Screenshot 2023-06-10 at 08 57 06


heckj:
😞 that Reddit has put us in this position, but makes sense.

Issue #4003, opened 2023-05-31
Search by author field not working


tylerbenson:
Some feeds are a combination from many authors. I want to be able to filter a feed to a specific author. I expected search to work, but this only applies to the content of the post, not the author.

For example, searching the [inessential](https://inessential.com/feed.json) feed for "Simmons" only shows one post.
![image](https://github.com/Ranchero-Software/NetNewsWire/assets/734411/39e59603-beaf-4e2f-b005-c401cdc1afea)

stuartbreckenridge:
I get the same results when testing with @brentsimmons' inessential.com feed. However, that's because—as you said—that feed has a top-level author and not a per-item author.

Testing with sixcolor's [feed](http://feedpress.me/sixcolors) and searching by author works for me. In the example below, I am searching for _Jason Snell_, and the first article has him as the author, and his name is not in the article text.

_All Articles_
<img width="2672" alt="Full Feed" src="https://github.com/Ranchero-Software/NetNewsWire/assets/7046652/8687d1a4-6fdc-4101-8f93-34e2e5394d58">

_Filtered_
<img width="2672" alt="Filtered for one Mr. Jason Snell" src="https://github.com/Ranchero-Software/NetNewsWire/assets/7046652/7ead39f9-3eeb-4024-b08c-f26b15b9ad07">

If you have feeds where things are potentially misbehaving, please let me know.

tylerbenson:
I'm not sure how to get the raw xml feed for that feed for comparison.

Here's an example feed with many different authors that doesn't allow filtering in this way:
https://notabird.site/i/lists/1348435279527333889/rss

stuartbreckenridge:
Ah, I think I see it.

@brentsimmons If I'm reading things correctly, we are only searching article content?

```swift
func fetchArticlesMatching(_ searchString: String) throws -> Set<Article> {
		var articles: Set<Article> = Set<Article>()
		var error: DatabaseError? = nil
		
		queue.runInDatabaseSync { (databaseResult) in
			switch databaseResult {
			case .success(let database):
				articles = self.fetchArticlesMatching(searchString, database)
			case .failure(let databaseError):
				error = databaseError
			}
		}

		if let error = error {
			throw(error)
		}
		return articles
	}
```

tylerbenson:
[...]

PR #4047, opened 2023-06-26
Add `authorID` to `search` virtual table

tylerbenson:
I'm not sure if this will solve #4003 as I don't know what `authorID` actually looks like.  It might need the `author.name` instead.


tylerbenson:
I guess I'll go ahead and close this PR and just leave #4003 open for someone that has more swift development skill than me.

PR #4158, opened 2023-11-22
Now set the correct base URL for each article's webview, and now load app JavaScripts as WebKit "user" scripts.

### LINKED ISSUE #4156: All images broken for https://blog.wadetregaskis.com
From looking at the RSS feed source (via `curl`) everything looks fine, and I can fetch the referenced images just fine.  The feed also works fine - images are shown correctly and reliably - in other RSS readers.  Only NetNewsWire (that I've tested) fails to load all the images, showing the broken image symbol.

URL:  https://blog.wadetregaskis.com/feed

wadetregaskis:
Setting the real base URL (rather than using a file URL pointing to the app's Resources folder) allows relative URLs to work correctly within the article, such as for images, and is compatible with Cross-Site-Origin policies that restrict use of resources outside of the origin domain.

It also implicitly eliminates access to the local file system from within the webview, as the use of a non-file base URL makes WebKit treats the webview's content as being from a remote server, and its default security policy is to then disallow local file access (except with explicit user action, such as drag-and-drop or via an `input` form element).

Note: the base URL is currently typically taken from the feed itself (specifically the "link" feed (channel) metadata).  That is controlled by the feed author (or a man-in-the-middle attacker).  It should perhaps be validated to ensure it's actually an HTTP/HTTPS URL, to prevent security problems.

The app-specific JavaScripts - used for fixing styling issues and the like - are now formally loaded as extensions to the web page, "user scripts" in WebKit parlance.  They're isolated to their own JavaScript world - meaning they can't be seen or manipulated by JavaScript from the feed article itself, and are more secure as a result.

Fixes #4156.

**Note**: the iOS target doesn't build at top of tree ("Type 'AddFeedIntentHandler' does not conform to protocol 'AddFeedIntentHandling'" and various other errors) so I haven't actually tested the iOS builds, just the Mac.

### DIFF
--- a/Mac/MainWindow/Detail/DetailWebViewController.swift
+++ b/Mac/MainWindow/Detail/DetailWebViewController.swift
@@ -96,6 +96,18 @@ protocol DetailWebViewControllerDelegate: AnyObject {
 		userContentController.add(self, name: MessageName.windowDidScroll)
 		userContentController.add(self, name: MessageName.mouseDidEnter)
 		userContentController.add(self, name: MessageName.mouseDidExit)
+
+		let baseURL = ArticleRenderer.page.baseURL
+		let appScriptsWorld = WKContentWorld.world(name: "NetNewsWire")
+		for fileName in ["main.js", "main_mac.js", "newsfoot.js"] {
+			userContentController.addUserScript(
+				.init(source: try! String(contentsOf: baseURL.appending(path: fileName,
+																		directoryHint: .notDirectory)),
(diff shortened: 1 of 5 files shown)


brentsimmons (approved):

brentsimmons:
Great work — thanks so much for this!

Wevah:
I'd totally forgotten we weren't using user scripts for our embedded scripts. Thanks for digging into this!

PR #4187, opened 2024-01-06
Add Spanish translation

martgnz:
Unfortunately I wasn't able to build main to test this patch. Ideally after main stabilises a bit we could create a build and test it.


martgnz:
Closing for now.

Issue #3606, opened 2022-06-18
Reproducible read-after-free crash when adding iCloud account


winniequinn:
After installing NetNewsWire on iOS and adding an iCloud account there, I installed 6.1 on macOS and attempted to do the same. Each time I attempt to do so, I immediately crash just as the "Use your iCloud account." dialog begins to show up.

Deleting all NetNewsWire data in "~/Library/Containers" and deleting all NetNewsWire data in iCloud does not resolve the problem. This is my first time ever attempting to use the application, so lingering data from a previous version cannot be an issue.

See [the crash report](https://github.com/Ranchero-Software/NetNewsWire/files/8934297/report.txt) for more info.

----

**Update:** It appears that I receive the same crash when attempting to add _any_ type of account.

vincode-io:
Thanks for the crash report. It looks like it might be some kind of SwiftUI bug.

Have you tried rebooting and then adding the iCloud account then? We've seen in recent versions of macOS that rebooting will sometimes clear up weird macOS bugs.

winniequinn:
@vincode-io: Restarting did not resolve the issue, I'm afraid.

winniequinn:
- No luck deleting everything in Containers.
- I'll try updating to 12.4 and will report back.
- I _do_ have iCloud Drive enabled, although I do have it _disabled_ specifically for "Desktop & Documents Folders". I would assume that's not the problem though.

winniequinn:
Thanks for trying to get to the bottom of this! I'll post again after I update to 12.4.

vincode-io:
I'm keeping my fingers crossed that the upgrade fixes it for you.

You shouldn't need the iCloud Drive folders stuff, so that isn't likely the problem.

winniequinn:
I'm getting the same crash with 12.4, unfortunately.

vincode-io:
The crashlogs looks the same then?

vincode-io:
Just to be clear, it is when you hit "Continue" in the below screenshot that you briefly see the dialog and then it crashes, right?

<img width="624" alt="Screen Shot 2022-06-20 at 2 38 23 PM" src="https://user-images.githubusercontent.com/16448027/174674877-6ac4b0c0-594c-4d56-9cdd-bd4a3190f405.png">

winniequinn:
That is correct. It crashes just as the dialog starts to fade in; I managed a screenshot after a few attempts:

[...]

PR #4244, opened 2024-03-20
Fix account add crash

winniequinn:
As reported in #3606, it was the case that attempting to add an account could result in a crash. This fixes the problem by retaining the `NSHostingController` that holds the `AddAccountsView` view in a manner similar to how 48138b1 retained `accountsAddLocalWindowController`.


winniequinn:
**NOTE:** This is a resubmission of #3670 which was merged to main by @vincode-io on September 18th, 2022. I see that this commit is no longer in main. (Perhaps because someone clobbered it with a force push?) I verified that this fix is still necessary and appropriate before resubmitting. Hopefully this be included in the next update!

Issue #3804, opened 2023-01-03
Mark All as Read in Confirmation


stuartbreckenridge:
The current Mark All as Read action sheet directs users to Settings to disable the confirmation dialogue. This appears to be an unnecessary redirect around the app. 

Would it be possible to streamline the options:

1. Mark All as Read in "X"
2. Mark All as Read in "X" and Disable This Confirmation

![image](https://user-images.githubusercontent.com/7046652/210294203-3746ea4b-6676-4b0c-a0d8-6a894c18eeac.jpeg)

vincode-io:
I can't remember for sure why we did it that way. It was probably so that if users wanted to reenable it, they knew how because we showed them how to shut it off.

kyvandoan:
I agree that in the action to the Settings view is unnecessary after it is first shown. 

An alternative to @stuartbreckenridge would be to automatically hide the `settingsAction` after it is first shown (in that case `message` should be slightly modified).

File Reference: [iOS/MasterTimeline/MarkAsReadAlertController.swift](https://github.com/Ranchero-Software/NetNewsWire/blob/0334885798606991d9033b8984c2db6bb0b84090/iOS/MasterTimeline/MarkAsReadAlertController.swift)

mtrolley:
I just came here to log this issue so I'm glad to see it's already requested. I agree with @kyvandoan that after it's shown once I don't think it needs to be shown again.

PR #4402, opened 2024-11-13
Hide Open Settings in Mark As Read confirmation

mtrolley:
Fixes issue #3804. After showing the note and quick link to Settings to disable the Mark As Read confirmation once don't show it again.

Tested on iPhone 16 with iOS 18.1 emulator.


mtrolley:
@brentsimmons Please let me know if you think this change is OK or would prefer it be addressed in another way.

PR #4430, opened 2024-12-11
Correct view controller identifier for iOS

kaiserabliz:
The word 'Web' is missing in the string that identifies VC in add storyboard


brentsimmons (changes requested):
I’d rather change the identifier in the storyboard to `AddFeedViewControllerNav`

kaiserabliz:
@brentsimmons should be good now

brentsimmons (approved):
👍

PR #4458, opened 2024-12-31
Beehiv fix

### LINKED ISSUE #4399: beehiiv blog showing metadata in preview
The blog, The Popehat Report, an example post at [https://www.popehat.com/p/god-bless-america](https://www.popehat.com/p/god-bless-america) isn't showing its preview properly. An example screenshot is attached, where you can note that the preview is a bunch of CSS-like metadata for beehiiv. 

I am also attaching the XML for the RSS, where you can easily see the XML pieces that have this metadata.

Screenshot 2024-11-11 at 14 30 17

[6QXMLvJf2y.txt](https://github.com/user-attachments/files/17708730/6QXMLvJf2y.txt)

stuartbreckenridge:
Fixes #4399 

Simulator Screenshot - iPhone 16 Pro Max - 2024-12-31 at 17 10 55


RSCore/Sources/RSCore/Shared/String+RSCore.swift
```diff
@@ -195,7 +195,15 @@ public extension String {
 	///
 	/// - Note: Doesn't work correctly with nested tags of the same name.
 	private func removingTagAndContents(_ tag: String) -> String {
-		return self.replacingOccurrences(of: "<\(tag).+?</\(tag)>", with: "", options: [.regularExpression, .caseInsensitive])
+		let pattern = "<\(tag)>.*?<\\/\(tag)>"
+		if let regex = try? NSRegularExpression(pattern: pattern, options: [.dotMatchesLineSeparators, .caseInsensitive]) {
+			let range = NSRange(location: 0, length: self.utf16.count)
+			let modifiedString = regex.stringByReplacingMatches(in: self, options: [], range: range, withTemplate: "")
+			return modifiedString
+		} else {
```
brentsimmons, line 203: Would you add a comment here explaining the approach? I’m no regex expert, and it would be helpful to me. (One question I have: wouldn’t there be a way to alter the pattern we were already using to include line separators?)

brentsimmons (changes requested):

Modules/Secrets/Sources/Secrets/SecretKey.swift
```diff
@@ -0,0 +1,64 @@
+// Generated by SecretKey.swift.gyb
```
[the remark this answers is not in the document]
stuartbreckenridge: Done.

RSCore/Sources/RSCore/Shared/String+RSCore.swift
stuartbreckenridge: Code comments added in d2364ff.

RSCore/Sources/RSCore/Shared/String+RSCore.swift
```diff
@@ -189,13 +189,37 @@ public extension String {
...
```
brentsimmons, line 219: Instead of trying one thing and then a second if it fails, could we just modify the original pattern — make it something like this?

```
<\(tag)>[\\s\\S]*?</\(tag)>
```

I’m not sure the above is totally right — but the idea is to look for zero or more of whitespace/non-whitespace between the tags, which should include line separators.

stuartbreckenridge: That approach also works. See 9b508e0.

No easy fix for nested tags though.

brentsimmons (approved):

Issue #4465, opened 2025-01-03
Feature Request: Register NetNewsWire to handle http and https, so Safari's Develop menu offers NetNewsWire in the Open Page With options.


faisal:
Bring back NetNewsWire classic's advertisement of being able to handle http: and https: URLs, which results in Safari adding it to the Develop > Open Page With menu.

This was addressed in https://github.com/Ranchero-Software/NetNewsWire/pull/3598, though that didn't come along for the rebase. I can't find discussion of a decision on whether to keep it, so I created a new issue in the hope of bringing it back.

Notes on why I think the option is worth keeping:
- The functionality is built in to the OS and rides on the OS's routing for URL schemes, so it's a little less fragile than the extension.
- To enable NetNewsWire's share extension in Safari is to enable the share extension all over the OS (Notes, Reminders, Numbers, Keynote, etc). That's a downside of the share sheet architecture as a whole.
- Extensions can be fragile, for reasons out of NNW's control.

PR #4466, opened 2025-01-03
Add http and https to the list of CFBundleURLSchemes.

faisal:
If NetNewsWire on macOS says it is able to handle http and https URL schemes then Safari will list NetNewsWire as a target in Develop > Open Page With, which provides a trivial flow for subscribing to the page or site one is currently reading in Safari. This has the advantage of providing a reasonably fast path for subscription without requiring users configure the app extension to get the functionality.

This change implements https://github.com/Ranchero-Software/NetNewsWire/issues/4465.

### DIFF
--- a/Mac/Resources/Info.plist
+++ b/Mac/Resources/Info.plist
@@ -37,6 +37,8 @@
 				<string>feed</string>
 				<string>feeds</string>
 				<string>x-netnewswire-feed</string>
+				<string>http</string>
+				<string>https</string>
 			</array>
 		</dict>
 	</array>


stuartbreckenridge:
You can use the Share Extension in Safari. Pinged you on Slack with instructions on how to enable - let me know if that satisfies the ask.

faisal:
I can use the Share Extension in Safari but it's a little problematic. I hinted at this in the rationale in https://github.com/Ranchero-Software/NetNewsWire/issues/4465.

brentsimmons (approved):

PR #4494, opened 2025-01-18
🩹 Fixes incorrect refresh control visibility 

### LINKED ISSUE #4490: Dual progress bars on iPad
IMG_0058

stuartbreckenridge:
Makes `TimelineViewController` a `UISplitViewControllerDelegate`.

- When the display mode is such that the feeds, timeline and article are visible, the timeline refresh control is hidden. 
- When the display mode is such that the timeline and article are visible, the timeline refresh control is visible.

Tested on iPad and iPhone 16 Pro Max (landscape).

Fixes #4490

### DIFF
--- a/iOS/MainTimeline/TimelineViewController.swift
+++ b/iOS/MainTimeline/TimelineViewController.swift
@@ -60,9 +60,6 @@ class TimelineViewController: UITableViewController, UndoableCommandRunner {
 		NotificationCenter.default.addObserver(self, selector: #selector(displayNameDidChange), name: .DisplayNameDidChange, object: nil)
 		NotificationCenter.default.addObserver(self, selector: #selector(willEnterForeground(_:)), name: UIApplication.willEnterForegroundNotification, object: nil)
 		
-		// Split View Controller Delegate
-		splitViewController?.delegate = self
-		
 		// Initialize Programmatic Buttons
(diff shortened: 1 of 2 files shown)


iOS/MainTimeline/TimelineViewController.swift
```diff
@@ -59,7 +59,10 @@ class TimelineViewController: UITableViewController, UndoableCommandRunner {
 		NotificationCenter.default.addObserver(self, selector: #selector(contentSizeCategoryDidChange), name: UIContentSizeCategory.didChangeNotification, object: nil)
 		NotificationCenter.default.addObserver(self, selector: #selector(displayNameDidChange), name: .DisplayNameDidChange, object: nil)
 		NotificationCenter.default.addObserver(self, selector: #selector(willEnterForeground(_:)), name: UIApplication.willEnterForegroundNotification, object: nil)
-
+		
+		// Split View Controller Delegate
+		splitViewController?.delegate = self
```
brentsimmons, line 64: The SceneCoordinator is the delegate for the splitViewController — so this fix should go in SceneCoordinator rather than in TimelineViewController

brentsimmons (changes requested):

iOS/MainTimeline/TimelineViewController.swift
stuartbreckenridge: Fixed in b5c592a

brentsimmons (approved):

Issue #3442, opened 2022-02-04
macOS: feature request - printing


DC01:
I would love to get printing functionality back.

Application: creating a PDF archive of dedicated web pages.

It used to be available with the old NetNewsWire app.
Will it come back ? I really love the UI and speed of the current app,
so getting back the ability to print would be really great.

Thanks,
Karsten

Wevah:
I've got code that makes this work, though there are a few things to sort out re: lazy-loading images.

brentsimmons:
I hate to disappoint, but it’s unlikely that we’ll add printing as a feature. I’ll leave this feature request open, just in case.

gurple:
Basic OS-native printing support would be very welcome.

I'd love to know the reason it's not on the map.

DC01:
I’m creating a tagged pdf archive from selected articles (scientific research)Von meinem iPhone gesendetAm 20.10.2023 um 08:14 schrieb gurple ***@***.***>:﻿
Basic OS-native printing support would be very welcome.
I'd love to know the reason it's not on the map.

PR #4621, opened 2025-05-04
Add print support for macOS and iOS

### LINKED ISSUE #3715: Print/Export to PDF
> @NetNewsWire I'm enjoying using NNW, but I'd like to be able to save the articles to PDF or printing to PDF (for reading on an e-ink reader). Exporting to PDF doesn't seem possible. Is this deliberate or an omission? Could it be added please? Thank you.

https://twitter.com/TommyRot10/status/1580125126989053952

LeoNatan:
This PR adds support for printing articles on macOS and iOS, using the standard print panels on each platform.

Screenshot 2025-05-04 at 20 57 52
Simulator Screenshot - iPhone 16 Pro Max - 2025-05-04 at 20 58 28

~Since articles' HTML body max width is constrained to 44em, a new web view is created using the HTML of the actual web view, its body is unconstrained and then a print is triggered from that web view.~
The stylesheet was modified to include printing-specific overrides, which remove padding, margin and turn link colors black.

Closes #3715 and #3442

### DIFF
--- a/Mac/Base.lproj/Main.storyboard
+++ b/Mac/Base.lproj/Main.storyboard
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="23504" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.Storyboard.XIB" version="3.0" toolsVersion="23727" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="23504"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="23727"/>
     </dependencies>
     <scenes>
         <!--Application-->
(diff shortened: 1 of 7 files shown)


LeoNatan:
Huh, apparently it is possible to have CSS styles only for printing. That’s should simplify a lot. I’ll modify the PR.

LeoNatan:
Hi @brentsimmons 
Any chance for a review?
Thanks in advance!

brentsimmons:
I wish you’d asked first! This is a dilemma for me. I don’t want to support printing in NetNewsWire.

But I do understand when people ask for ways to export articles. Ideally we’d support several export formats (HTML, JSON, PDF, perhaps others).

How about this: would you revise this PR to add (only on Mac), an `Export as PDF…` command? It should be in the File menu, in its own section, below the Subscriptions commands.

LeoNatan:
May I ask, any particular reason you do not wish to support printing? It is still something we do, even if less commonly and less frequently. I saw your comments on previous issues, but assumed the core team just doesn't want to do the work to support it. It turned out to be very little work.
But sure, I can add the option to export to a PDF. I think it's possible to have for both macOS and iOS, but I can limit to macOS only. It will go through the same mechanism as the printing for creating the PDF.

brentsimmons:
NetNewsWire has never supported printing in all of its 23 years. I think of it as a modern application, web-based, forward-looking, and not looking back to the ’80s and ’90s when printing was commonplace.

There are two other reasons:

* I am personally responsible for maintaining, testing, and providing customer support for all features, and printing can be a difficult feature to support. (Different printers, different ideas of how an article should print, etc.) NetNewsWire users will benefit more from my spending my time on things other than printing
* In almost all cases, you can open the original article in a browser and print it. NetNewsWire is not stopping anymore from printing an article they want to print.

LeoNatan:
Yes, but one of the advantages of NNW is the clean presentation of text. Even with reader modes, the result is limited at best when compared to dedicated reader apps. And as you can see from the PR code itself, the support surface of this functionality is very narrow; we use WebKit's printing capabilities and some minor adjustments to the CSS.

I am willing to take the role of "maintainer" of print functionality, such as it is.

brentsimmons:
I very much appreciate your generous offer! But I’ve learned that I can’t count on anyone besides myself to be around and active indefinitely. That’s not meant personally! It’s just the way it is. Which means I have to be personally willing to do 100% of the maintenance and customer support.

And: while I understand your point about text presentation, I’ve never wanted to add a Print command to NetNewsWire and I still don’t.

LeoNatan:
Cheers, so let’s close the PR.

PR #4661, opened 2025-06-13
Early Work on Liquid Glass

stuartbreckenridge:
This commit focuses on initial iOS 26 changes.

What's included: 

### 🔧 Config
- Deployment targets are updated to 26.0

### 🚨 Compiler fixes
- Make `Transport` type conform to `Sendable`
- Adds `override` keyword to `toggleSidebar`

### 🚑️ Bug fixes
- Certain code moved from `viewDidLoad` to `viewWillAppear` [d774e53] to workaround crashes.

### 💄 User Interface
- Rebuilt in Xcode 26 (obviously)
- Main feeds now uses `insetGrouped`. 
- Refresh progress and updates are now communicated via the navigation view subtitle. 
- The search bar is now in the bottom toolbar on iPhone and in the top right of the article view on iPad.
- The Settings "Done" button moved from the left to the right to have platform consistency.
- The Filter button now follow same look and behaviour as Mail.
- The bottom toolbar in the article view has been grouped as follows:
    - Left: Article actions (mark as read, mark as favourite)
    - Middle: Next Unread
    - Right: Other actions (reader view, share)
- The Mark All as Read `svg` has been tweaked to vertically centre the image so that it has the appropriate positioning in a circular button. 

### 💩 Known Issues
- VibrantTableViewCell is still used on Feeds and Timeline. This leaves some visual artefacts.
- Timelines selection on iPad extends all the way under the sidebar. 
    - Hilariously, you have to pull waaaaay to the right to reveal the leading-edge swipe action on the cell.
- Search scopes don't appear on iPad.
- State restoration doesn't restore refresh progress. 
- Filter button on iPad moves left/right when active/inactive.

Simulator Screenshot - iPad Pro 13-inch (M4) - 2025-06-13 at 20 32 47

### DIFF
--- a/iOS/Settings/Settings.storyboard
+++ b/iOS/Settings/Settings.storyboard
@@ -500,7 +500,7 @@
                         </connections>
                     </tableView>
                     <navigationItem key="navigationItem" title="Settings" id="uf8-rU-NlH">
-                        <barButtonItem key="rightBarButtonItem" style="done" systemItem="done" id="HlH-s0-OwW">
+                        <barButtonItem key="rightBarButtonItem" systemItem="cancel" id="HlH-s0-OwW">
                             <connections>
                                 <action selector="done:" destination="a0p-rk-skQ" id="Mzh-25-h49"/>
                             </connections>


iOS/MainFeed/MainFeedViewController.swift
```diff
@@ -699,6 +699,38 @@ class MainFeedViewController: UITableViewController, UndoableCommandRunner {
...
+				}
+			} else {
+				navigationController?.navigationBar.topItem?.subtitle = ""
+			}
+
+		} else {
+			navigationController?.navigationBar.topItem?.subtitle = NSLocalizedString("Updating...", comment: "Updating...")
+		}
+		
+		scheduleNavigationBarSubtitleUpdate()
```
brentsimmons, line 725: Would you add a check so that it doesn’t schedule an update when an update is already scheduled? Or maybe just make this a timer scheduled for once a minute? (As written, if there are several calls to updateNavigationBarSubtitle in a row for whatever reason, then it would schedule multiple updates.)

This can be done in a follow-up PR, of course.

brentsimmons (approved):

PR #4663, opened 2025-06-14
Liquid Glass: Fixes to Nav Bar Subtitle; macOS 26 build; and config files.

stuartbreckenridge:
## Navigation Bar Subtitle

Subtitle logic has been moved to the SceneCoordinator to keep everything in sync and in one place. The logic for what subtitles display follows Mail and is included in code commentary. 

The unread count badge has been removed from the navigation bar title view and constraints updated.

Known issue: iPad navigation subtitle is left aligned while the title view is centre aligned.

## macOS

A small change to constraints was required. NNW in macOS 26 is otherwise up-and-running.


iOS/SceneCoordinator.swift
```diff
@@ -539,6 +549,94 @@ class SceneCoordinator: NSObject, UndoableCommandRunner {
...
+			// Updating in progress, apply to both iPhone and iPad Feeds.
+			self.mainFeedViewController?.navigationItem.subtitle = NSLocalizedString("Updating...", comment: "Updating...")
+		}
+		
+		scheduleNavigationBarSubtitleUpdate()
+		
+	}
+	
+	func scheduleNavigationBarSubtitleUpdate() {
+		if isRefreshScheduled {
```
brentsimmons, line 631: I’m thinking `isRefreshScheduled` sounds like it’s related to whether or not a feeds-refresh is scheduled.

Maybe this should be renamed to something like `isNavigationBarSubtitleUpdateScheduled`? That’s a long name, but more accurate. (I think — unless I misunderstand the code.)

Could be done in a follow-up PR, of course.

brentsimmons (approved):

PR #4750, opened 2025-09-11
Fixes for unread counts, device names, and disabling of JS on iDevices

stuartbreckenridge:


brentsimmons (changes requested):
Looks good! Just a few changes requested.

For the future, let’s separate multiple issues into multiple PRs. (It’s easier and quicker for me to review.)

iOS/AppDefaults.swift
```diff
@@ -189,6 +190,15 @@ final class AppDefaults {
...
 	
+	var disableJavaScript: Bool {
+		get {
+			return AppDefaults.bool(for: Key.disableJavaScript)
+		}
+		set {
+			AppDefaults.setBool(for: Key.disableJavaScript, newValue)
+		}
+	}
+
```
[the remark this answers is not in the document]
stuartbreckenridge: Done

iOS/Article/PreloadedWebView.swift
```diff
@@ -46,6 +48,11 @@ class PreloadedWebView: WKWebView {
 		}
 	}
 	
+	@objc func userDefaultsDidChange(_ sender: Any) {
+		configuration.defaultWebpagePreferences.allowsContentJavaScript = AppDefaults.shared.isArticleContentJavascriptEnabled
+		reload()
+	}
```
brentsimmons, line 54: This will result in the page being reloaded any time any UserDefaults value changes — which can be surprisingly often. See how the Mac version is handling this, to make sure it reloads only when the JavaScript setting changes.

brentsimmons (changes requested):
One more change requested — how we handle reloading when the setting changes.

iOS/Article/PreloadedWebView.swift
stuartbreckenridge: Fixed in 402fbd3.

brentsimmons (approved):

PR #4878, opened 2025-12-12
Update Sepia style

meyerweb:
Tune up the link styling in Sepia to further improve accessibility


brentsimmons:
According to GitHub, there aren't actually any changes — maybe still need to add a file?

meyerweb:
Yeah, I… have absolutely no idea how I managed to convince git to commit a no-file-no-edits “change” and then open a PR with it.  I’ll close this one and see if I can do it right the next time.  Sorry for the noise!

PR #4930, opened 2026-01-08
🚑️ Use NSSharingServicePicker to resolve crash

### LINKED ISSUE #4929: A Control-click on a timeline entry crashes v7
This code causes the crash:

```swift
let standardServices = NSSharingService.sharingServices(forItems_noDeprecationWarning: items) as? [NSSharingService] ?? [NSSharingService]()
```

stuartbreckenridge:
Fixes #4929

- Always show a "Share..." menu, but disable it when necessary.
- `Share...` presents `NSSharingServicePicker` (similar to Toolbar)
- The `NSSharingServicePicker` is anchored to the trailing x coordinate of the selected row

### DIFF
--- a/Mac/MainWindow/Timeline/TimelineViewController+ContextualMenus.swift
+++ b/Mac/MainWindow/Timeline/TimelineViewController+ContextualMenus.swift
@@ -190,41 +190,31 @@ private extension TimelineViewController {
 			}
 		}
 
-		if let sharingMenu = shareMenu(for: articles) {
-			menu.addSeparatorIfNeeded()
-			let menuItem = NSMenuItem(title: sharingMenu.title, action: nil, keyEquivalent: "")
-			menuItem.submenu = sharingMenu
-			menu.addItem(menuItem)
-		}
+		menu.addSeparatorIfNeeded()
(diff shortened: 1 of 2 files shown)


Mac/MainWindow/Timeline/TimelineViewController+ContextualMenus.swift
```diff
@@ -190,41 +190,31 @@ private extension TimelineViewController {
...
 		}
 
-		if let sharingMenu = shareMenu(for: articles) {
-			menu.addSeparatorIfNeeded()
-			let menuItem = NSMenuItem(title: sharingMenu.title, action: nil, keyEquivalent: "")
-			menuItem.submenu = sharingMenu
-			menu.addItem(menuItem)
-		}
+		menu.addSeparatorIfNeeded()
+		let shareButtonMenuItem = NSMenuItem(title: NSLocalizedString("Share...", comment: "Share..."), action: #selector(showShareSheet(_:)), keyEquivalent: "")
```
brentsimmons: Style nit: instead of three dots ... we should use a single ellipsis character … which you can type using option-;

PR #5020, opened 2026-01-25
✨ Timeline as CollectionView

### LINKED ISSUE #5015: iOS: Selecting an article in timeline causes unread indicator to animate out then back in
The item is still marked read (note bottom toolbar in article view), but the indicator goes away and then comes back.

Image

### LINKED ISSUE #5019: Convert Timeline to UICollectionView

stuartbreckenridge:
This PR converts the Timeline to a `UICollectionView` and will close #5019. 

There are a few differences to note:
- Swipe actions—e.g., `toggleStar`—have a small delay applied in order to allow swipe animations to complete. Otherwise, the app tries to reload them mid-animation and a warning appears in the log.
- Swipe behaviours match Mail.
- Long-press menus now behave across iPhone and iPad.
- Fixes #5015 

Smaller items to be resolved:
- Leading swipe actions have incorrect offset on iPad when the sidebar is presented.

### DIFF
--- a/iOS/MainTimeline/MainTimelineModernViewController.swift
+++ b/iOS/MainTimeline/MainTimelineModernViewController.swift
@@ -509,7 +509,7 @@ extension MainTimelineModernViewController: UICollectionViewDelegate {
 		
 		let previewView = cell.contentView
 		let parameters = UIPreviewParameters()
-		parameters.backgroundColor = .white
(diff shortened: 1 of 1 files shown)


ashishgandhi:
Very excited! (Perhaps more than I should be.)

Issue #4793, opened 2025-09-24
Feature Request - Copy link/share with highlight


Babypopdip:
Feature Request
Would be awesome to see “Copy link with highlight” added to the text selection menu like safari has below! Or the option to do the same thing from the share sheet like safari also does! 

<img width="660" height="1434" alt="Image" src="https://github.com/user-attachments/assets/ce09672c-39ea-4751-bd1b-12a2ca8ae5f9" />

totalsoda:
I believe enabling Text Fragments would implement this the same as Safari and would be very nice so +1

Text Fragments act like:

[This](https://www.macrumors.com/how-to/share-webpage-link-quoted-text-messages-ios/#:~:text=Keep%20reading%20to%20learn%20more)

(I think)

PR #5068, opened 2026-02-01
Add Copy Link with Highlight feature for iOS

jpgmeyer:
Addresses #4793

## Summary
Adds a "Copy Link with Highlight" option to the context menu on iOS. When text is selected in an article, this option copies a [text fragment URL](https://developer.mozilla.org/en-US/docs/Web/URI/Fragment/Text_fragments) that will highlight the selected text when opened in a browser, similar to existing macOS version behaviour.

Capture d’écran 2026-02-01 à 15 06 51

## Changes
- New `TextFragmentURL` utility for creating text fragment URLs
- Added `getSelectedText()` JavaScript function to retrieve selected text from the WebView
- Added "Copy Link with Highlight" to the WebView context menu
- The standard share sheet is unchanged and always uses the plain article URL, to keep the scope narrow.

### DIFF
--- a/Shared/Article Rendering/main.js
+++ b/Shared/Article Rendering/main.js
@@ -136,6 +136,12 @@ function error() {
 	document.body.innerHTML = "error";
 }
 
+// Returns the currently selected text in the WebView, or empty string if no selection
+function getSelectedText() {
+	const selection = window.getSelection();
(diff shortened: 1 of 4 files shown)


brentsimmons:
@jpgmeyer Thanks! I’ll get to reviewing this once we ship 7.0 (any day now)

PR #5108, opened 2026-02-13
Fix style of italic and emphasized text in articleTitle

justin-b-reeves:
`body a *` is overriding the color of `<i>` and `<em>` in `articleTitle`.

```
body a, body a:visited, body a :not(i, em) * {
		color: var(--secondary-accent-color);
	}
```

Also works.

(“Hyperlegible” is the only preinstalled theme that doesn’t do this.)

### DIFF
--- a/Shared/Article Rendering/stylesheet.css
+++ b/Shared/Article Rendering/stylesheet.css
@@ -421,6 +421,10 @@ a.footnote:hover,
 		color: var(--secondary-accent-color);
 	}
 
+	a :is(i, em) {
+		color: inherit;
+	}
+
 	.externalLink a {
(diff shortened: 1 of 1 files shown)


brentsimmons:
I’m not clear on what the bug is and what the fix is — what does it look like when it's wrong?

justin-b-reeves:
Before:

Screenshot 2026-02-16 at 4 52 44 PM

https://justin-reeves.net/example.xml

justin-b-reeves:
After:

Screenshot 2026-02-16 at 4 57 32 PM

brentsimmons (approved):

Issue #853, opened 2019-08-15
Feature request: sort subscriptions by unread count


brentsimmons:

lilyball:
What I'd like to see is an option to sort all feeds with unread items to the top, but within unread feeds, use the normal sorting.

brentsimmons:
Related https://github.com/Ranchero-Software/NetNewsWire/issues/356

PR #5149, opened 2026-02-24
Add sidebar feed sorting options (by name, by unread count)

willie:
Adds a "Sort Feeds By" submenu to the View menu on macOS with sort type (Name, Unread Count) and direction (Ascending, Descending) options. The preference persists across launches. iOS is unchanged — it always sorts alphabetically ascending.

Switching sort type resets the direction to that type's natural default — Name sorts A–Z, Unread Count sorts most-unread first — like clicking a different column header. Re-selecting the current sort type leaves the direction alone.

When sorted by unread count, the sidebar keeps its order as counts change while reading; it re-sorts on the next tree rebuild (feeds added/renamed, sort changed, relaunch). This is deliberate — re-sorting live would shuffle rows under the pointer as articles are read.

Addresses #853 for Mac.


willie:
Updated the PR to match the current app conventions.

PR #5177, opened 2026-03-09
Enable Designed for iPad support on Apple Vision Pro.

### LINKED ISSUE #4832: Feature Request: visionOS Checkbox
I'm not talking about adding the target support, but just allowing iPad versions. Apple Vision Pro could be great for reading RSS.

I will be happy to do proper visionOS APP too.

bbhoss:
Fixes #4832

This change allows the Apple Vision Pro to run NetNewsWire in iPad mode. See the [video](https://youtu.be/F03wW4eu8_o), it performs pretty well I think. Obviously it would be best if the embedded browser worked as expected, but personally I'd much rather have the app in this form than waiting. This is the app I use for reading RSS across all my Apple devices and not having access to it on AVP is a let down. If there's anything else I can do to get this change in, please let me know.


ErikAbele:
I'll happily second that. It would be awesome to have NetNewsWire at least in "Compatibility Mode" on the AVP.

Nobody expects to have official support for this mode (especially on the AVP, most of the power-users there are devs anyway) and a note in the FAQ could clarify that it is just offered as-is.

In my experience most of the other apps which are available as iPad apps on the AVP do actually work quite fine, except for Electron apps which just display a web UI in a wrapper app (they do work but have tiny UIs impossible to navigate with your eyes & hands).

But this is not the case here. It would work as it does on iPad.

Issue #553, opened 2019-02-21
Some buttons in YouTube embeds don't work


sdtransier:
Version 5.0d15 (1517)

If there is a YouTube embed in the feed, the following buttons/links don't work:

- Main title
- Watch Later
- "YouTube" button on bottom bar
- Fullscreen

I don't know if this is a YouTube embed limitation or NNW issue. 

<img width="659" alt="screen shot 2019-02-21 at 7 50 15 am" src="https://user-images.githubusercontent.com/17955542/53177538-78929f00-35ad-11e9-8209-655b3a337750.png">

brentsimmons:
What’s the URL of the feed where you’re seeing this?

sdtransier:
It's the most recent post in http://feeds.kottke.org/json

brentsimmons:
I tested — none of my code in NetNewsWire gets called on clicks on any of these things.

joshuajabbour:
Yes, I’m noticing the same issue. Always happens on kottke.org YouTube embeds, though I have no issues on the videos when viewed directly on his site. I found a video embedded in another NNW default feed (the iBeer post on furbo.org), which also exhibits the issue. 

I just spent a while debugging this further, and was able to reproduce in other feed readers. News explorer displays the embeds inline like NNW, and also has the issue. Neither Reeder 3 nor Unread 1 display embedded videos inline, so they don’t have the issue. 

Therefore I assume it’s an issue with YouTube’s embed code in some way. Unfortunate, because it’s super annoying to not be able to open videos directly in the YouTube app.

jimtut:
Not sure I agree w/ the last comment.  My feeds include MacRumors and Dark Horizons, and I've used the old NNW (3 or 4?) on a very old Mac without problems w/ this.  Upgrading to Big Sur and NNW 5 with the same feeds gives me this problem.

I've tried w/ the same feeds saved "locally" and with them stored in Feedly.  Definitely a problem w/ the Mac desktop version, haven't tried much w/ the iOS version, but I think it's the same.

brentsimmons:
If those worked in old NetNewsWire versions, it’s almost certainly because we were using the old WebView instead of the new WKWebView, which replaces it. They’re similar, yes, but different enough to cause a variety of issues.

brentsimmons:
Can't reproduce with the 7.1.3 development version. Closing this. We can reopen after 7.1.3 ships if it’s still an issue.

Issue #4860, opened 2025-12-02
Embedded YouTube player doesn't work (Error 153)


That-Human-Being:
Basically whenever a feed item has a YouTube video attached to it like how most hackaday articles link to the YouTube video that the article is about, the embedded player shows up, but is unable to play the video (I've tried adjusting various settings but none of what I tried did anything)

![image](https://github.com/user-attachments/assets/a5bf979a-0ddd-446c-b795-1204dc819a62)

[The article the screenshot was taken of](https://hackaday.com/2025/12/01/necroprinting-isnt-as-bad-as-it-sounds/)

brentsimmons:
What’s the URL of the feed? Which syncing system are you using? (iCloud, Feedbin, or…?)

That-Human-Being:
URL is https://hackaday.com/blog/feed/
And currently I'm just syncing to "on my iPhone"

johan456789:
https://developers.google.com/youtube/terms/required-minimum-functionality#embedded-player-api-client-identity

> API Clients that use the YouTube embedded player (including the YouTube IFrame Player API) must provide identification through the HTTP Referer request header. In some environments, the browser will automatically set HTTP Referer, and API Clients need only ensure they are not setting the [Referrer-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Referrer-Policy) in a way that suppresses the Referer value. YouTube recommends using strict-origin-when-cross-origin Referrer-Policy, which is already the default in many browsers.

This should fix it: https://stackoverflow.com/a/79805072/6306190

> You need to either add this header:
'Referrer-Policy': 'strict-origin-when-cross-origin'
> Or you can add the following to your embed element:
referrerpolicy='strict-origin-when-cross-origin'

regeniaguenthurwag11-dot:
I encountered the same issue as well, the embedded player is not working. Perhaps Apple's native player could be an alternative option? I hope this issue gets fixed soon.

404erreurs:
Same problem here when using iCloud, all videos included in the articles have “error 153.”

regeniaguenthurwag11-dot:
The surprising news is that NNW running on Mac has been updated to version 7.0, which brings visual updates to the liquid glass effect. However, it seems that support for embedded playback in YouTube is still not available. Given the nature of NNW's development, I believe this major update may still take a long time......

made2k:
[...]

Issue #3206, opened 2021-06-28
Feature Request: macOS - Easy way to go full screen for embedded YouTube content


bwalls:
The one thing I miss from the Feedly app when reading feeds on my Mac is the ability to simply play a YouTube video full screen when I'm reading an article with an embed. There, if you click on the YouTube logo in the player, it will open the video's page at YouTube.com in the default browser, maintaining your place in the video (though often only after a commercial). From there the full-screen button will work.

Being able to just click full screen in NetNewsWire directly would be great, but I get that that is difficult. 

In NetNewsWire, I typically will click on the title to go to the article, possibly scroll to the video I'm interested in, and either go full screen from the embedded player in the browser or go from there to YouTube. I see I can right-click on the NetNewsWire-embedded player and select "Copy video URL at the current time," navigate to my browser, paste in the YouTube URL, and then go full screen. Is there a better way?

bwalls:
Year-and-a-half later I came back to post the same request, having completely forgotten this original post. Now for fullscreen I click on the YouTube logo to launch in Safari, where I normally have to pay for full screen with a commercial or two.

vordenken:
I know this is a quite old but for me clicking on the Youtube logo doesn't do anything, so there's now (comfortable) way to open the video in safari. 
Is this a known limitation/bug @brentsimmons or should I maybe open a new bug report?

PR #5211, opened 2026-03-24
Allow WKWebView to go fullscreen

jsonfry:
To support things like YouTube video embeds going fullscreen.

The built in behavoiur for multiple windows / monitors is pretty good, here's a video of it in action:

https://github.com/user-attachments/assets/17064a39-848b-4997-bf62-aa52e396020c

### DIFF
--- a/Shared/Article Rendering/WebViewConfiguration.swift
+++ b/Shared/Article Rendering/WebViewConfiguration.swift
@@ -36,6 +36,7 @@ private extension WebViewConfiguration {
 		let preferences = WKPreferences()
 		preferences.javaScriptCanOpenWindowsAutomatically = false
 		preferences.minimumFontSize = 12
+		preferences.isElementFullscreenEnabled = true
 
 		return preferences
 	}


jsonfry:
This possibly resolves #553 and #3206

brentsimmons:
Are there any risks to this? Side effects or edge cases?

What happens when a person doesn't have spaces turned on and it goes fullscreen? Is it obvious how to get out of fullscreen?

jsonfry:
Thanks for looking at this, I really appreciate it.

In terms obviousness of getting out of fullscreen, it depends on the contents of the webpage that has asked to go fullscreen, but you can always move the mouse pointer to the top of the screen:

Screenshot 2026-03-27 at 16 05 33

I think side effects and risks are minimal as WKWebView handles it all for you, it should behave the same as Safari. E.g. you see the same "Click to Exit Full Screen" behaviour in Safari:

Screenshot 2026-03-27 at 16 18 24

Also, given the built-in web rules, content can't go fullscreen unless the action happens from direct user input (like a click), so there's the same risk level as a web browser in terms of malicious web content.

I've also tested what happens if you have content in fullscreen and then switch to another feed or entry, and that behaves correctly.

I couldn't work out how to turn off spaces in Sequoia or Tahoe (the two Macs I have here), I'd hope that WKWebView would use whatever OS fullscreen api is available or report to the web content that fullscreen is not available, but that could be an interesting thing to check if anyone else can?

This is a feed I've checked: https://www.henrikkarlsson.xyz/feed . The entry "Differently Free" has an embedded YouTube video that works (quite why that embed doesn't have the "Error 153" problem in #4860 I don't know).

PR #5224, opened 2026-03-31
Fix default browser detection using malformed URL NSWorkspace.urlForA…

### LINKED ISSUE #5220: MacOS app doesn't correctly detect system default browser
My system default browser is set to Firefox, but I had to set that manually in NetNewsWire, which still thinks that Safari was the default browser:
Image

I tried restarting my computer after changing the system default, but that didn't help NetNewsWire find it.

MacOS Sequoia version 15.6.1
NetNewsWire version 7.0.3

Thanks for the incredible free app!

umutkonmus:
Fix default browser detection using malformed URL NSWorkspace.urlForApplication(toOpen:) was called with "https:///"

(malformed, no host), which caused macOS to sometimes return Safari regardless of the actual system default browser.

Fixed by using a valid HTTPS URL (https://apple.com/) so the system correctly resolves the default handler.

Also fixed the browser popup selection logic to explicitly handle the nil-defaultBrowserID case, and removed force-unwraps on icon/name.

Fixes #5220

### DIFF
--- a/Mac/Preferences/General/GeneralPrefencesViewController.swift
+++ b/Mac/Preferences/General/GeneralPrefencesViewController.swift
@@ -142,9 +142,9 @@ private extension GeneralPreferencesViewController {
 		let defaultBrowser = MacWebBrowser.default
 
 		let defaultBrowserFormat = NSLocalizedString("System Default (%@)", comment: "Default browser item title format")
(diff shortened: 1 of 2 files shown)


brentsimmons:
Closing because I believe this was fixed by a separate commit.

PR #5327, opened 2026-06-13
Find Feed: prefer a page-specific feed over a site-wide <head> feed (#5299)

### LINKED ISSUE #5299: Find feed for Relay blog finds different feed
From feedback:

Tried to add the RSS feed for this blog: https://www.relay.fm/blog

NetNewsWire added this RSS feed: http://relay.fm/master/feed

It works correctly if I manually add the RSS feed that they indicated on their website: https://www.relay.fm/blog/feed

Not sure if this is a NetNewsWire bug or just an issue with how they set up the blog.

bojanstef:
Fixes #5299. I asked first on Discourse and got the go-ahead: https://discourse.netnewswire.com/t/first-time-contributing-can-i-fix-issue-5299/299

### Problem
Find Feed on `https://www.relay.fm/blog` selects `http://relay.fm/master/feed` (the all-network feed) instead of the page's own `https://www.relay.fm/blog/feed`.

### Root cause
The page advertises two feeds:
- `http://relay.fm/master/feed` — a `<link rel="alternate">` in `<head>` (`.HTMLHead`)
- `https://www.relay.fm/blog/feed` — a body "Subscribe via RSS" link (`.HTMLLink`)

In `FeedFinder.findFeedsInHTMLPage`, as soon as any `<head>` feed is found it sets `didFindFeedInHTMLHead = true` and returns the `<head>` feeds, so the body candidate sitting in `feedSpecifiersToDownload` is discarded. `bestFeed` then receives a single-element set and returns `master/feed` without ever scoring — so `/blog/feed` is dropped before it can be considered. (Confirmed by tracing the discovery at runtime: both feeds were discovered; the body feed was discarded by the early return.) This isn't a relay.fm misconfiguration — both feeds are valid and advertised.

### Fix
Before falling back to the `<head>` feed, validate any body-linked candidate whose URL is on the same host and nested under the requested page's path (e.g. `/blog/feed` for a `/blog` request) and prefer it. Host comparison is `www`- and case-insensitive (via `localeForLowercasing`). Root-domain requests (no path) keep the existing `<head>`-feed behavior, so whole-site Find Feed is unchanged.

### Testing
- New unit test `testFeedURLIsUnderRequestedPagePath` — on-path preference, `www`/case-insensitive host matching, different-host rejection, root-path exclusion, and the sibling-prefix trap (`/blogger` is not under `/blog`).
- `swiftlint lint --strict` clean; `FeedFinder` module build has no new warnings; module tests pass.
- Verified end to end: Find Feed on `relay.fm/blog` now resolves to `https://www.relay.fm/blog/feed`.

### Disclosure
Diagnosed and drafted with LLM assistance (Claude Opus 4.8), reviewed to meet the project's quality bar — per CONTRIBUTING.md.

### DIFF
--- a/Modules/FeedFinder/Sources/FeedFinder/FeedFinder.swift
+++ b/Modules/FeedFinder/Sources/FeedFinder/FeedFinder.swift
@@ -144,6 +144,32 @@ public final class FeedFinder {
 	}
 }
 
+extension FeedFinder {
+
+	/// True when `feedURLString` is on the same host as `pageURLString` and its
+	/// path equals or is nested under the page's path. The page must have a
+	/// non-root path — a root request (e.g. example.com/) gets no on-path
+	/// preference, preserving the existing <head>-feed behavior for whole sites.
+	static func feedURLString(_ feedURLString: String, isUnderRequestedPageURLString pageURLString: String) -> Bool {
+		guard let feedURL = URL(string: feedURLString), let pageURL = URL(string: pageURLString),
+			let feedHost = feedURL.host(), let pageHost = pageURL.host() else {
+			return false
+		}
+		func normalizedHost(_ host: String) -> String {
+			let lowered = host.lowercased(with: localeForLowercasing)
+			return lowered.hasPrefix("www.") ? String(lowered.dropFirst("www.".count)) : lowered
+		}
+		guard normalizedHost(feedHost) == normalizedHost(pageHost) else {
+			return false
+		}
+		let pagePath = pageURL.path.hasSuffix("/") ? String(pageURL.path.dropLast()) : pageURL.path
+		guard pagePath.count > 1 else {
+			return false
+		}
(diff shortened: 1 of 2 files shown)


bojanstef:
@brentsimmons There's AI-generated comments, let me know if I should:
- [ ] Leave them
- [ ] Make them more concise 
- [ ] Remove them completely

bojanstef:
@brentsimmons would it help if I sent a screen recording?

brentsimmons:
I’m not sure this is actually the fix I want — we might just be trading a fix for this one specific site for breaking other sites. Not sure what to do yet. It may be just that it needs a special case for this particular site.

bojanstef:
@brentsimmons fair enough, let me know if you want to brainstorm
thanks

brentsimmons:
I ended up fixing this with a special case for the Relay blog.

Issue #5284, opened 2026-05-06
[Feature] Enclosure caching


thefranke:
Some feeds provide enclosures or Media RSS content. A famous example is from LBRY/Odysee, which adds MP4s of uploaded videos directly into the feed.

Here is an example of the latest item/entry in the [3Blue1Brown RSS feed](https://odysee.com/$/rss/@3Blue1Brown:b):

```xml
<item>
<title>Covering 10 points, a surprisingly tricky puzzle.</title>
<description>
<p><img src="https://thumbnails.lbry.com/QLu_ZsRc_G0" width="480" alt="thumbnail" title="Covering 10 points, a surprisingly tricky puzzle." /></p>Made as part of a monthly series of puzzles for the 2026 Year of Math.<br />...<br />https://www.youtube.com/watch?v=QLu_ZsRc_G0
</description>
<link>
https://odysee.com/covering-10-points%2C-a-surprisingly:312234cb2a834bc85cf6f2e586f989c314cc7651
</link>
<guid isPermaLink="true">
https://odysee.com/covering-10-points%2C-a-surprisingly:312234cb2a834bc85cf6f2e586f989c314cc7651
</guid>
<pubDate>Thu, 16 Apr 2026 16:08:21 GMT</pubDate>
<enclosure url="https://player.odycdn.com/api/v3/streams/free/covering-10-points,-a-surprisingly/312234cb2a834bc85cf6f2e586f989c314cc7651/8ffc3a.mp4" length="1420267" type="video/mp4"/>
<itunes:title>Covering 10 points, a surprisingly tricky puzzle.</itunes:title>
<itunes:author>3Blue1Brown</itunes:author>
<itunes:image href="https://thumbnails.lbry.com/QLu_ZsRc_G0"/>
<itunes:duration>38</itunes:duration>
<itunes:explicit>false</itunes:explicit>
</item>
```

The use case I describe now is a regular one for travelers: You sit an an airport or travel in a country with very irregular connections and only strafe a Wifi here and there. The RSS reader could have a workflow for starred items, that when enabled downloads and caches all enclosures on the device until the item is un-starred. In addition, the reader could allow you to either open the file via URL handlers or with a built-in viewer (as do other readers have).

This way, NetNewsWire would act as a kind of offline viewer not just for downloaded RSS feed descriptions, but for enclosures as well. If I enter an airplane, I can watch all my recently starred videos, open PDFs or listen to audio files, completely offline.

[...]

Issue #3460, opened 2022-02-11
Unified Image Handling


vincode-io:
There are several areas regarding image handling that could be improved and enhanced. I think unified image architecture for fetching and caching images could address most of these.

### Favicons

I think could do something more sophisticated than completely clearing the icon caches. Maybe we periodically check to see if the image has changed when it is accessed. Only if it has been changed would we update it and notify the UI to reload. This should remove the flashing of the default Globe for feeds with Favicons/Web Icons.

### Image Zooming

I'd also like to change how we load article images to go through an internal cache of our own. We could use that image cache to look up images when zooming them. This would get rid of the current delay we have on iOS. It would also be a good time to add image zooming to the Mac as well as make the iOS app zooming more sophisticated.

See: 

https://github.com/Ranchero-Software/NetNewsWire/issues/647
https://github.com/Ranchero-Software/NetNewsWire/issues/3349
https://github.com/Ranchero-Software/NetNewsWire/issues/3354
https://github.com/Ranchero-Software/NetNewsWire/issues/3360

### Timeline Hero Image Thumbnail & Offline Viewing

This internal cache could be optionally loaded during the sync process. Just off the top of my head, I see 3 different modes. No caching on sync, only cache the hero image thumbnail, and full caching. Caching the hero image would let us use it quickly in the Timeline. Full caching would be used for people who want to use NNW without internet access (train commuters, etc...).

See: 

https://github.com/Ranchero-Software/NetNewsWire/issues/3302
https://github.com/Ranchero-Software/NetNewsWire/issues/1873

### Implementation Details

The heart of this work would be a replacement for `ImageDownloader`. It might be tempting to also try to replace some of the classes that use `ImageDownloader`, but there is some complicated, hard to unpack logic in those classes and they work. Let's leave those largely intact.

We would need to parse the article body to find the images and identify the hero image. The article body would also need to be changed to support a new NetNewsWire only protocol for images so that we could run them through a custom protocol handler that utilizes the `ImageDownloader` replacement.

The sync process would need to be enhanced to cache images depending on user preference. This might be a good time to enhance our progress indicator to show textual detail about what it is doing. That way people don't feel like they have to keep open the application when all we may be doing is trying to cache images.

Article zooming code would need to be rewritten on iOS and implemented on macOS.

vincode-io:
We should evaluate [Nuke](https://github.com/kean/Nuke) with our usual skepticism to see if it makes sense include some or all of it as a dependency.

[...]

Issue #1873, opened 2020-03-10
Request for iOS: Cache images for off-line reading


vangeliis:
So so good to see you back on iOS NetNewsWire!

Please can you add the option to cache images for off-line reading?

Thanks,
Vangeis

brentsimmons:
When are you reading off-line? I’m curious about the use case.

vangeliis:
Hi Brent,

Many thanks for getting back to me.
I live in London and read my feeds while commuting with the tube. We don’t have wifi in the tunnels so having everything cached on the phone makes it so much easier.
What do you think?

Regards,
Vangelis

vincode-io:
I want this one quite frequently.  I’ve often wanted to load up on articles while I’ve had good service, knowing that I won’t have service in transit or at my destination.

If you live on one of the US coasts, you rarely have a problem getting data cellular service. In the middle of America it is much more spotty. My brother lives in a suburb of Omaha, NE and I can’t get data at his house. Lots of times office buildings block your cellular data too and you might not be able to put a personal device on the company WiFi. 

I would also like to cache Reader View article downloads for the same reason as images.

scrosland:
Like @vangeliis I often want to cache articles before going on the tube (London underground), and also on long train journeys across country when I know that there will be spotty cellular data coverage. With Reeder I would open the app and wait for the articles and images to be cached before setting off.

I think it's also worth considering that cellular data is expensive in many parts of the world.

brentsimmons:
This sounds like a good idea. Let’s remember this when we’re planning iOS 5.2.

PorridgePi:
Hi,
I would really appreciate it too if this feature could be added into NetNewsWire. I have a limited data plan and could not afford  to use my data to view all the images while on commute everyday, so I would usually download or cache some articles. Really hope that this feature would be added! Thanks.

grizzee:
I'm also very interested in this feature. I like to read articles on an iPad and often in trains and in other enviroments with no wifi. Hope this will be implemented in the near feature.

matt17r:
It looks like this is already tentatively planned but I’d like to add my plus one. My use case is for reading in cellular and wifi dead zones.

Before finding this issue I googled (DDG’d) to see if it’s possible and found [a thread on Reddit](https://www.reddit.com/r/rss/comments/y4og8j/netnewswire_any_way_to_download_images/)  asking the same question so there’s a few of us wanting this ability.

tylerbenson:
[...]

PR #5352, opened 2026-07-02
Add offline article-image caching behind a setting

davidfwatson:
Adds an optional, off-by-default setting that caches article images during refresh and
serves them from a local cache at render time, so articles read with images while offline
(planes, trains, tunnels, metered data). Addresses #1873; a step toward #3460.

## Design note: what happens when you turn it on

Prefetch only catches images from *future* refreshes, so the moment you enable this —
typically right before you lose connectivity — your *existing* unread articles still have
no cached images. To bridge that, **enabling the setting offers to cache images for your
current unread articles right away** (a "Cache Images Now? / Not Now" prompt showing the
unread count). There's deliberately no standalone "cache everything" button: if you want
to backfill again later, toggle the setting off and back on.

image

image

## How it works

Two halves sharing an on-disk cache:

- **Prefetch:** `ArticleImagePrefetcher` watches `.AccountDidDownloadArticles`, extracts
  `` URLs off the main thread (a sync can deliver a large batch at once), and fetches
  them via `ArticleImageDownloader` — throttled, not one unbounded burst. The pure
  URL-extraction is a nonisolated, unit-tested `ArticleImageURLExtractor`.
- **Serve:** with the setting on, `main.js` rewrites `http(s)` `` to
  `nnwimage://cache/?u=…`; `ArticleImageSchemeHandler` (a `WKURLSchemeHandler`) resolves
  those memory → disk → network. Uncached images load online-only; cached ones load
  offline. On rewrite it also drops the img's `srcset` and, inside a `<picture>`, its
  sibling `<source>` candidates, so WebKit can't select an original `https` URL that
  bypasses the cache. The enable flag reaches `main.js` as a hidden DOM marker rather than
  an inline `<script>`, so it applies even when the reader has article JavaScript off (that
  setting gates *content* JS; `main.js` runs as an injected user script).

`ArticleImageDownloader` mirrors the existing `ImageDownloader`/`FaviconDownloader` pattern
(same `Downloader`, `ImageMetadataDatabase` backoff, `BinaryDiskCache`) but keeps full-size
bytes in a separate `ArticleImages` folder. Same-URL requests coalesce; the in-memory
`NSCache` is bounded (~50 MB, cleared on background/low-memory); only plausible image
responses persist (an `image/*` type or recognized magic bytes, under a 20 MB cap). App-side
fetches reuse the web view's content blocker via `ArticleImageContentBlocker`, so tracker/ad
URLs stay blocked — including on cross-host redirects. The `ArticleImages` folder is swept by
the existing 3-day `CacheCleaner` flush, and purged outright when the setting is off.

## The setting

`cacheImagesForOffline`, off by default — iOS: Settings → Articles; Mac: Settings → General.
While a backfill runs, the settings line shows live progress ("Caching Images… N of M");
when idle it shows the current cache size ("166 images · 63.5 MB used"). iOS account-stats
gains an Offline Images section with the same figures.

## Key limitations

- Custom-scheme routing bypasses WebKit's own image cache and doesn't send its cookies/referer,
  so a few hotlink-protected images may fail when the setting is on.
- Enclosure/media caching (video, audio, PDF) is out of scope (#5284).
- No pre-buffer byte cap: the 20 MB limit keeps oversized bodies out of the *persistent* cache,
  but the shared `RSWeb.Downloader` still buffers whole bodies first. A streaming cap belongs in
  `Downloader` (it affects every image path) and is left as shared-infra follow-up.

## Testing

- macOS and iOS both build clean.
- `ArticleImageURLExtractorTests` covers absolute / relative / protocol-relative resolution,
  scheme lowercasing, `data-canonical-src` preference, entity decoding, non-http rejection,
  dedup, ordering.
- Manual: enable → accept the prompt → Airplane Mode → confirm article images still load.

*(Full internals live in the three commit messages: core caching, on-enable backfill, and the
Mac prefs-pane sizing fix.)*

### DIFF
--- a/Mac/AppDefaults.swift
+++ b/Mac/AppDefaults.swift
@@ -43,6 +43,7 @@ final class AppDefaults: Sendable {
 		static let defaultBrowserID = "defaultBrowserID"
 		static let currentThemeName = "currentThemeName"
 		static let articleContentJavascriptEnabled = "articleContentJavascriptEnabled"
+		static let cacheImagesForOffline = "cacheImagesForOffline"
 
 		// Hidden prefs
 		static let showDebugMenu = "ShowDebugMenu"
@@ -319,6 +320,15 @@ final class AppDefaults: Sendable {
 		}
 	}
 
+	var cacheImagesForOffline: Bool {
+		get {
+			UserDefaults.standard.bool(forKey: Key.cacheImagesForOffline)
+		}
+		set {
+			UserDefaults.standard.set(newValue, forKey: Key.cacheImagesForOffline)
+		}
+	}
+
 	init() {
 		// Migrate every-10-minute refresh interval to 30 minutes.
 		let rawValue = UserDefaults.standard.integer(forKey: Key.refreshInterval)
@@ -344,7 +354,8 @@ final class AppDefaults: Sendable {
 			Key.refreshInterval: RefreshInterval.every2Hours.rawValue,
 			Key.showDebugMenu: showDebugMenu,
 			Key.currentThemeName: Self.defaultThemeName,
-			Key.articleContentJavascriptEnabled: true
+			Key.articleContentJavascriptEnabled: true,
+			Key.cacheImagesForOffline: false
 		]
 
 		UserDefaults.standard.register(defaults: defaults)
--- a/Mac/AppDelegate.swift
+++ b/Mac/AppDelegate.swift
@@ -147,6 +147,7 @@ let appName = "NetNewsWire"
 		installAppleEventHandlers()
 
 		CacheCleaner.purgeIfNecessary()
+		ArticleImagePrefetcher.shared.start()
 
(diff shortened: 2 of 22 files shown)


davidfwatson:
Rebased onto the latest `main` (now at `808403b00`).

Since this PR was opened, `main` refactored `RSWeb.Downloader.download(_:)` to return a `DownloadResponse` struct instead of a `(Data?, URLResponse?)` tuple. The rebase merged cleanly textually but broke the build at that API boundary, so I updated `ArticleImageDownloader.performDownload` to the new signature — matching the sibling `ImageDownloader` / `SingleFaviconDownloader` call sites. No behavioral change. Both the iOS and macOS targets archive clean again.

davidfwatson:
Not sure if there's anything more I can do to help this along. Let me know if there's some reason this is not a useful addition.

Issue #5361, opened 2026-07-09
Substack style footnotes aren't detected


dave-atx:
NetNewsWire currently doesn't detect the way Substack inserts footnotes into articles. For [an article like this one][1], the footnotes just show up as plain links, then open an embedded WebView to the article if they're clicked on. You don't get any of the fancy popover handling like you do with MultiMarkdown style ones.

Screenshot of the iOS app (v7.1.1 TestFlight) below. Behavior is the same on macOS.

<img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/8618c204-e889-46a0-8de5-d59d33aaab35" />

[1]: https://www.natesilver.net/p/america-used-to-be-exceptionally

PR #5362, opened 2026-07-09
Support Substack-style footnotes in the article popover

dave-atx:
This addresses issue #5361 .

"Substack-style" footnote references are bare anchor tags instead of being wrapped in `<sup>` tags like MultiMarkdown. This PR adds detection for that style of footnote to the Javascript in the Article Rendering stack so that they render with the fancy popovers as well.

### DIFF
--- a/Shared/Article Rendering/main.js
+++ b/Shared/Article Rendering/main.js
@@ -138,11 +138,13 @@ function error() {
 
 // Takes into account absoluting of URLs.
 function isLocalFootnote(target) {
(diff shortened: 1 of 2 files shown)


dave-atx:
FYI, I've added substack-style and a bunch of additional common footnote formats to a [theme that I use][1]. I'm happy to amend this PR to pull in those in as well. It would affect the same files that the PR currently touches, but just add to the diff.

[1]: https://github.com/dave-atx/ember-nnw-theme/blob/main/test/footnote-providers.toml

PR #5399, opened 2026-08-23
feat(reader-api): add custom HTTP headers for self-hosted accounts

### LINKED ISSUE #3920: Allow custom HTTP headers when connecting to FreshRSS
It is common for self-hosted servers to have an additional layer of authentication to access services. Most of these authentication services allow for access via token in an HTTP header, thus avoiding the direct exposure of APIs. It would be nice to be able to set custom headers when setting up connection to self-hosted FreshRSS instances.

jeankhawand:
closes #3920

FreshRSS-customheaders-mac
FreshRSS-customheaders


jeankhawand:
Hello @brentsimmons please let me know if there is anything needed to get this merged. Thanks

